PR #543, opened 2014-12-05
Upgrade to version 7

gislenius:


dknochen:
@gislenius there seem to be 2 problems with the api:
1. facttable-for-report.jq doesn't return any result (click on Facts tab)
2. spreadsheet-for-report.jq returns duplicate facts for each synonym (click on spreadsheet tab -> e.g. Assets contains 2 facts in one cell)

you can reproduce the behaviour using the following simple report:
https://dl.dropboxusercontent.com/u/1074594/salkdfalkfja/test.xbrlb

could you look into these issues?

PR #580, opened 2015-01-06
Fix hints

fcavalieri:


fcavalieri:
Hi @dknochen @gislenius , 
can any of you review this when you have time? This fixes the hints introducing support for different hints according to the deployment profile. I use the profile stored in $config:profile-name even if the query overrides it. Thanks

fcavalieri:
Closed as the target branch is now merged

PR #583, opened 2015-01-07
Mark sec specific parameters

wcandillon:


dknochen:
yes, I agree. Can we make it arrays? Also the japan branch merge introduces a new endpoint in the swagger file that needs adaptation as well.

dknochen:
the cellstore profile is called sec. That's what we should use I think. secxbrl is the deployment.

dknochen:
after renaming secxbrl -> sec we can merge.

wcandillon:
@dknochen make sense thks!

PR #584, opened 2015-01-07
Add API examples

wcandillon:


wcandillon:
@dknochen This is ready for review, I'm currently adding/fixing all examples. But the infrastructure is here.

wcandillon:
@dknochen Ready for review.
List of future improvements with high priority:
* e2e tests
* support for arbitrary parameters and arrays of parameters

app/api/api.html
```diff
@@ -0,0 +1,64 @@
...
+            <div class="scrollable">
+                <h2><span ng-bind="op.nickname"></span>()</h2>
+                <p ng-bind="op.summary"></p>
+                <pre ng-bind="curl()"></pre>
+                <div ng-if="examples.length > 0">
+                    <h3>Examples</h3>
+                    <ul>
+                        <li ng-repeat="example in examples">
+                            <a ng-click="test(example.parameters)" ng-bind="example.title"></a>
+                        </li>
```
dknochen: I think we need to make the examples more prominent similar to this:

image

dknochen: Also the first example should be selected by default. Starting from an empty form is not so straight forward because you don't know what to put into the fields anyway.

app/api/api.html
```diff
@@ -0,0 +1,64 @@
...
+                                <td class="parameter"><input type="text" ng-model="params[param.name]"></td>
+                                <td ng-bind-html="param.description"></td>
+                            </tr>
+                        </table>
+                    </div>
+                </div>
+                <button ng-click="test()" class="btn btn-primary" ng-class="{ disabled: loading }">
+                    <span ng-if="!loading">Test</span>
+                    <span ng-if="loading"><i class="fa fa-spin fa-spinner"></i> Loading...</span>
+                </button>
```
dknochen: the Test button deserves an icon I think

app/reports/reports.html
```diff
@@ -2,14 +2,20 @@
...
             </div>
             <account></account>
         </div>
     </header>
     <div class="container scrollable std-top-padding">
+        <h1>APIs</h1>
         <section>
-            <button class="btn btn-primary" title="Create a New Report" ng-click="createReport()" id="create-report">CREATE</button>
+            <button class="btn btn-primary" title="API Documentation" ui-sref="api({ method: 'POST', path: 'entities.jq' })"><i class="fa fa-wrench"></i> API Documentation</button>
+        </section>
```
dknochen: Maybe this is not important right now, but given that we will add more admin pages we could already think about structuring the layout accordingly. It could become messy if we add more pages like this.

app/api/api.html
```diff
@@ -0,0 +1,64 @@
...
+                    </ul>
+                </div>
+                <div ng-init="isCollapsed = false">
+                    <h3><a class="fa" ng-class="{ 'fa-caret-down': !isCollapsed, 'fa-caret-right': isCollapsed }" ng-click="isCollapsed = !isCollapsed"></a> Parameters</h3>
+                    <div collapse="isCollapsed">
+                        <table class="table">
+                            <tr ng-repeat="param in op.parameters">
+                                <td ng-bind="param.name"></td>
+                                <td class="parameter"><input type="text" ng-model="params[param.name]"></td>
+                                <td ng-bind-html="param.description"></td>
```
dknochen: it would be absolutely awesome if we could also list the options in case of an enum, e.g. list "json", "csv", "xml", "excel" for:

                        {
                            "name": "format",
                            "description": "The result format",
                            "required": false,
                            "type": "string",
                            "enum": [
                                "json",
                                "csv",
                                "xml",
                                "excel"
                            ],
                            "paramType": "query"
                        }

app/api/api.html
```diff
@@ -0,0 +1,64 @@
...
+                    </li>
+                </ul>
+            </div>
+        </block-container>
+        <block-splitter></block-splitter>
+        <block-container>
+            <div class="scrollable">
+                <h2><span ng-bind="op.nickname"></span>()</h2>
+                <p ng-bind="op.summary"></p>
+                <pre ng-bind="curl()"></pre>
```
dknochen: Here something comes to my mind about the token. The token used should not be the ui token and it should be selectable from a dropdown (only the app tokens should be selectable). This will become a security issue when you want to share the urls with others because the ui-token will allow the recipient to login to the ui.

dknochen:
How can we define different enums for different profiles in the swagger file? for example:

    {
        "name": "tag",
        "description": "A tag to filter",
        "profile": ["secxbrl"],
        "required": false,
        "type": "string",
        "enum": [
            "DOW30",
            "SP500",
            "RUSSEL1000",
            "FORTUNE100",
            "PJI",
            "ALL"
        ],
        "paramType": "query"
    }

and: 

    {
        "name": "tag",
        "description": "A tag to filter",
        "profile": ["edinet"],
        "required": false,
        "type": "string",
        "enum": [
            "NIKKEI",
            "ALL"
        ],
        "paramType": "query"
    }

dknochen:
The "Headers" and "Tests" Tab are empty:

image

dknochen:
only nice to have: syntax highlighting for the response body ;)

dknochen:
very fancy stuff

PR #597, opened 2015-01-09
Make cik parameter non sec specific

### LINKED ISSUE #591: Typo in Japan example?
In `report-elements.jq`, `cik` is documented as sec only.
However the edinet example is as:
http://edinet.28.io/v1/_queries/public/api/report-elements.jq?_method=POST&token=54eb0587-1aee-44a5-bb56-627f6ef43da6&cik=E01225&format=json&profile-name=japan&label=%22Loss%20of%20inactive%20facilities%22

Is it a typo where it should eid instead of cik?

wcandillon:
fixes #591

### DIFF
--- a/swagger/queries.json
+++ b/swagger/queries.json
@@ -56,7 +56,6 @@
                         {
                             "name": "cik",
                             "description": "A CIK number",
-                            "profile": ["sec"],
                             "required": false,
                             "type": "string",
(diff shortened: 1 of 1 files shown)


gislenius:
I believe that it should not be made generic. See other pull request that adds it to the Japanese profile.

PR #602, opened 2015-01-11
Minor improvement for new config files

wcandillon:


dknochen:
we will create a lot of copies. But for now it should be ok.

wcandillon:
@dknochen nice catch! Indeed that doesn't scale :)

wcandillon:
@dknochen I now erase the previous backup, what do you think?

dknochen:
@wcandillon perfect

PR #605, opened 2015-01-12
added waiting for bucket to exist.

dknochen:
fixing these errors:

image

### DIFF
--- a/tasks/s3.js
+++ b/tasks/s3.js
@@ -89,6 +89,22 @@ var listObjects = function (idempotent, prefix, marker, contents) {
     return defered.promise;
 };
 
+var waitForBucketExists = function() {
+    var defered = Q.defer();
+    s3.waitFor('bucketExists', {
+        Bucket : bucketName
+    }, function(err) {
+        if (err) {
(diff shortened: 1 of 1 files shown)


dknochen:
@wcandillon ready for review

tasks/s3.js
```diff
@@ -96,11 +112,17 @@ var createBucket = function() {
         ACL : 'public-read'
     }, function(err, data) {
         if (err || data === null) {
-            $.util.log(bucketName + err);
+            $.util.log($.util.colors.red(bucketName + err));
             defered.reject();
         } else {
             $.util.log('createBucket(' + bucketName + ')');
-            defered.resolve();
+            waitForBucketExists()
```
wcandillon: `return waitForBucketExists()`

wcandillon:
@dknochen Looks good but this instabilities appears since we switched to docker. I'm wondering if this is linked. Also each build now creates a lot of bucket we need to clean up maybe.

dknochen:
@wcandillon whenever it worked so far it was pure luck because the createBucket is asynchronous. So, it is a race condition in any case. With docker it became faster and the race condition started to show as failure.

PR #615, opened 2015-01-13
Add git commit message guidelines

wcandillon:


wcandillon:
@dknochen What do you think?

dknochen:
@wcandillon it is certainly both professional and useful. And, - after thinking about it - I now do understand the rational behind using present tense. I will forget to use the emoticons though, but lets see.

PR #618, opened 2015-01-14
codegeneration with mustache

dknochen:


app/app.js
```diff
@@ -2,6 +2,7 @@
 
 angular.module('report-editor', [
     'lodash',
+    'Mustache',
```
wcandillon: @dknochen Why introduce a new dependency? Lodash seems to do the job and to be even faster: http://jsperf.com/mustache-compliant-libraries/18

wcandillon:
@dknochen Looks good except for the mustache dependency that might not be required. If you get rid of Mustache too from the devDependencies (gulp) that would be :tada:

app/index.html
```diff
@@ -2,6 +2,7 @@
 <html ng-app="report-editor" resize>
 
     <head>
+        <base href="/" />
```
wcandillon: @dknochen That's an unrelated change isn't it? I would like to discuss test this change separately.

PR #620, opened 2015-01-15
fix missing base element error

dknochen:


wcandillon:
@dknochen Can you add a test that wouldn't work without this change? I removed the base uri for a reason but I forgot why. I need to investigate.

dknochen:
I did a clean reinstall and the error disappeared :fire:

PR #629, opened 2015-01-19
Fix tag handling in Japan profile.

gislenius:


fcavalieri:
@gislenius Hi Ghislain, where is $tags defined?

fcavalieri:
@gislenius Hi Ghislain, I updated the indexes and relaunched the tests.

PR #630, opened 2015-01-20
Switching to TDNET+EDINET

gislenius:


gislenius:
@fcavalieri I believe that this is good to go ;-)

gislenius:
Also, note the dependency on https://github.com/28msec/cellstore/pull/631 as ignoring array order is required for the tests to pass with 7_3.

fcavalieri:
Hi @gislenius, good to go as soon as the bug on #631 is fixed.

PR #631, opened 2015-01-20
Add option to ignore array ordering in expected test results.

gislenius:


gislenius:
In report element tests, the ordering of objects in array keeps changing when the database is repopulated. Since the order is irrelevant, I added an option to ignore it. That way we don't need to constantly update the expected results.

fcavalieri:
Hi @gislenius, two arrays which contain the same unique elements and have the same number of items will be deemed equal even if they are not.

See: http://try.zorba.io/queries/xquery/bFe58n6%2BWBfKMGESH32eI6u9yPA%3D

PR #633, opened 2015-01-20
Catch mongo query errors

dknochen:


fcavalieri:
I think that if the query contains a streamable non seekable string the error would become "[ZSTR0055] Streamable string already consumed" as the query is consumed when sent to mongo and consumed again by the serialize call in the error handler you wrote. Nonetheless, I don't think this will happen as streamable non seekable strings should not be in the query. It is fine for me. What do you think?

dknochen:
@fcavalieri Hey federico, I removed consuming the query again as it is a security issue anyway.
Is it better now?

fcavalieri:
I love it, just remove the $hint parameter so that the test pass and you
can merge it

PR #636, opened 2015-01-21
Add labels.jq endpoint.

gislenius:


queries/modules/io/28/modules/xbrl/components.jq
```diff
@@ -91,6 +111,33 @@ declare function components:components($component-or-ids as item*) as object*
...
+  $roles as string*,
+  $concepts as string*,
+  $exact-labels as string*,
+  $options as object*
+) as object*
+{
+  let $aids := if(deep-equal($archives-or-ids, $components:ALL-ARCHIVES))
+               then $components:ALL-ARCHIVES
+               else archives:aid($archives-or-ids)
+  let $query := {|
```
dknochen: could we throw an error if all of the parameters are set to $components:ALL-*?

queries/modules/io/28/modules/xbrl/components.jq
```diff
@@ -56,6 +56,26 @@ declare variable $components:ROLE as xs:string := "Role";
 declare variable $components:DEFAULT-LANGUAGE as xs:string := "Language";
 
 (:~
+ : Placeholder for all archive IDs..
+ :)
+declare variable $components:ALL-ARCHIVES as xs:string := "";
```
dknochen: I really think we should not use an empty string here. It is hard to see the semantics of it when you debug. Could we use "ALL" instead? That would be even more consistent.

dknochen:
@gislenius 2 simple comments only. Thank you!

queries/modules/io/28/modules/xbrl/components.jq
gislenius: DONE

queries/modules/io/28/modules/xbrl/components.jq
gislenius: DONE

queries/modules/io/28/modules/xbrl/components.jq
```diff
@@ -91,6 +111,43 @@ declare function components:components($component-or-ids as item*) as object*
...
+  |}
+  return
+    switch(true)
+    case $roles eq $components:ALL-ROLES and
+         $aids eq $components:ALL-ARCHIVES and
+         $concepts eq $components:ALL-CONCEPTS and
+         $exact-labels eq $components:ALL-LABELS
+      return error(
+        QName("components:TWO-MANY-COMPONENTS"),
+        "Two many components to be returned because no filtering is done."
```
dknochen: typo: two -> too and TWO -> TOO ;)

gislenius: DONE ;-)

wcandillon:
This is so cool. Can I redeploy edinet?

PR #641, opened 2015-01-21
Add /facttable-for-component.jq examples

wcandillon:
fixes 28msec/cellstore-nttdata#15


gislenius:
Pending Dennis' fix for multiple parameters value so that the example can really show multiple companies and multiple years.

PR #645, opened 2015-01-22
Migrate to japanxbrl-7_4.

gislenius:


dknochen:
@gislenius good to go :)

PR #646, opened 2015-01-23
Test examples

dknochen:


dknochen:
@wcandillon can you take a look and give me feedback if you think it should work this or another way. Once, we settled the general framework, we can add tests for all examples in the queries.json and make it even mandatory.

What do you think?

swagger/queries.json
```diff
@@ -369,6 +369,14 @@
                     "description": "Get list of all periods.",
                     "parameters": {
                         "tag": "ALL"
+                    },
+                    "test": {
+                        "japan": [
+                            {
+                                "transform": "($response.Statistics.NumPeriods)",
+                                "expectedResult": 27
```
wcandillon: @dknochen I would like to keep the API documentation with the database decoupled.

dknochen: @wcandillon It makes sense to me. Though this argument also applies to the other branch. And, the test results are also coupled with the deployments.

What if we would remove the test fields before publishing this file in dist/swagger? This way we could keep it clean and seperate... Somehow I like to test both - backend and frontend. The backend testing can be more fine grained data related testing.

What do you think?

wcandillon: maybe this should go in queries/public/test/edinet/*

wcandillon:
@dknochen I will try to revive this. It seems like it could be incredibly useful in the case of the MarkLogic branch.

PR #650, opened 2015-01-26
fix dependencies of 28 tasks

dknochen:


wcandillon:
@dknochen I'm investigating, it has to be related with the change I've made with the config overwrite. Maybe I shouldn't have merge that.

PR #651, opened 2015-01-26
Components fixes

gislenius:


wcandillon:
@gislenius add example in the swagger file.

queries/modules/io/28/modules/xbrl/components.jq
```diff
@@ -905,3 +905,64 @@ declare %private function components:line-items-recursive($components as object*
           case $substitution-group eq "xbrldt:hypercubeItem" return ()
           default return $concept-name
 };
+
+declare function components:hypercubes($components as object*) as string*
```
dknochen: could we add documentation for these 3 functions?

gislenius:
@dknochen good point, back to you! :-)

PR #652, opened 2015-01-26
Test example e2e

dknochen:


dknochen:
@wcandillon could you take a look? This now adds autogenerated e2e tests from the swagger file. Thanks!

wcandillon:
@dknochen Looks good but it seems to be merged with your other PR. Also it looks like there are conflicts. Could you clean this up to have only the e2e part?

PR #653, opened 2015-01-26
Revert bogus configuration change

wcandillon:


dknochen:
@wcandillon do you understand why this has caused the problem?

PR #654, opened 2015-01-26
Add onlyTextBlocks option

wcandillon:


wcandillon:
@gislenius This is my attempt :)
I hope you can take it from there. It looks like there is a dependency with the question you asked by email regarding how `IsTextBlock` computed.

Could you also point me out how to ask for report-elements across all entities?

wcandillon:
@gislenius This attempt might be completely bogus, feel free to ignore it if it doesn't make sense ;-)

gislenius:
Hi William,

It was the right idea, but took a bit of more work to put it at the right place.

I also added a link to the "dimensionless" facts associated with a report element within its archive.

Kind regards,
Ghislain

queries/public/api/report-elements.jq
```diff
@@ -1,6 +1,7 @@
 import module namespace api = "http://apps.28.io/api";
 
 import module namespace session = "http://apps.28.io/session";
+import module namespace backend = "http://apps.28.io/test";
```
dknochen: this module should never be used in the api.

queries/public/api/report-elements.jq
```diff
@@ -222,9 +215,18 @@ let $result :=
                           Language: $language,
                           Value: $concept.Labels.$labelRole.$language
                         }
-                      ]
+                      ],
+                      Fact: backend:url("facts", {
```
dknochen: This will disclose the test user token. We should refactor this out into a seperate module. Then use the new module's function in the test module and here (returning the url containing the current users token).

queries/public/api/report-elements.jq
```diff
@@ -222,9 +215,18 @@ let $result :=
...
+                          "xbrl:Concept": $original-name,
+                          aid: $archive,
+                          fiscalYear: "ALL",
+                          fiscalPeriod: "ALL",
+                          fiscalPeriodType: "ALL",
```
dknochen: should we skip those 3 for profile-name eq "generic"? Should we also add a Fact field for the sec profile?

queries/public/api/report-elements.jq
dknochen: Also, we should maybe rename the field to Facts (with an s)

gislenius: Actually, there should only be one single fact since it's a single archive and dimensionless. Several facts would mean a collision, hence I did not put an s. Does it make sense?

gislenius: I created a new module. That the token appears in the output is on purpose. It's the whole point of navigating through the API. This is not a breach: you only get the token if you supplied it in the call. Does it make sense?

dknochen: thanks for addressing this. The point was not that a token is in the output, the point was that the token of the test user is in the output, which still needs to be fixed:
https://github.com/28msec/cellstore/commit/6dd0b65ec20c95f2c8da028fbf63a229a3600467#diff-8ce11d58abe55bdfc61294f5fd8d7760R31

1. in the tests the test user's token should be used
2. for any other user, a token of that user should be used

The test user token is only for the tests and should never be exposed. Does it make sense?

gislenius:
@dknochen There you go! :-)

dknochen:
@gislenius you changed the order of the parameters and that fixed the tests?

gislenius:
Hi Dennis,

Yes. The tests are sensitive to the order (deep-equal on strings) and the order changed because of my last edit.

Kind regards,
Ghislain

dknochen:
@gislenius Got it! Makes sense. Thank you.

PR #656, opened 2015-01-27
Merging entities

gislenius:


gislenius:
The tests are failing because of the missing taxonomy in the DB. They should work again as soon as the data is here.

queries/public/api/components.jq
```diff
@@ -90,15 +92,25 @@ let $tag as string* := api:preprocess-tags($tag)
 let $reportElement := ($reportElement, $concept)
 let $networkIdentifier := distinct-values(($networkIdentifier, $role))
 
-(: Object resolution :)
+let $cik1 as string* := switch($profile-name)
+            case "sec" return $cik
+            case "japan" return $edinetcode
+            default return ()
+
+let $cik2 as string* := switch($profile-name)
```
dknochen: why do we add a $cik2 parameter and not use the $ticker (which is the securities code)?

dknochen:
@gislenius the securitiescode is still documented in the swagger docs. We could keep it as an alias for ticker in the API. What do you think?

gislenius:
@dknochen I now fixed the doc.
I think an additional parameter is too much noise and would like to keep entropy low. I already feel uncomfortable not reusing the cik one, but that's SEC terminology.
Does it make sense?

dknochen:
good to go :). Thank you.

PR #662, opened 2015-01-29
Remove entity _id field in output for Japan profile.

gislenius:


gislenius:
In the Japanese profile, entities have multiple EIDs and the _id field is a generic UUID that we should not expose and that will break at each new population of the cell store.

PR #663, opened 2015-01-29
Add documentation for onlyTextBlocks and fix for SEC.

gislenius:


gislenius:
Filtering text blocks was not implemented in the SEC profile, and was not documented either.

PR #665, opened 2015-01-29
Replace array with link to labels in report elements (except SEC profile).

gislenius:


gislenius:
Since we now have a labels endpoint, we can flatten the report-elements endpoint and link to the label endpoint.

PR #668, opened 2015-02-03
Fix entity retrieval in facts.jq.

gislenius:
Fixes 28msec/cellstore-nttdata#25


wcandillon:
@gislenius does it fix https://github.com/28msec/cellstore-nttdata/issues/25? If yes, can we had an example?

PR #669, opened 2015-02-03
Fix test on master.

gislenius:


gislenius:
The test is now ignoring the order in which components are output.

PR #681, opened 2015-02-06
Dashboard

wcandillon:


wcandillon:
@dknochen Right now it doesn't work for the sec profile but I would like to merge this anyways. It is not linked from anywhere.

PR #683, opened 2015-02-06
Abstract filtering

wcandillon:


wcandillon:
Would it make sense to add `onlyTextBlocks` and `abstract` to `labels.jq`?

gislenius:
Hi William,

I will later do a major improvement of report element classification which should involve these parameters as well. I would rather polish then and stick to what corresponds to needed use cases.

Kind regards,
Ghislain

gislenius:
Hi @wcandillon,

If you agree, that's good to go.

Kind regards,
Ghislain

wcandillon:
@gislenius Right now the use case is to have it on `labels.jq` because this what we use in the dashboard. But I agree that this has low priority.

PR #694, opened 2015-02-09
Switch to larger japan database

wcandillon:


wcandillon:
deprecated by japanese deployment.

PR #695, opened 2015-02-10
Switch to japanxbrl-8_2.

gislenius:


wcandillon:
@gislenius Can you take a look? The examples added by @dknochen don't seem to work against this db.

PR #704, opened 2015-02-12
Entity full-text search

wcandillon:


wcandillon:
@gislenius Can you review the backend part please?

wcandillon:
* multiplexer:entities function
* Move the documentation of search below token
* Document parameter as `Includes in the results the entities whose name match the search term (5 characters minimum).`
* Instead of  either-or, would be better to do an and.

wcandillon:
- Make search parameter consistent

wcandillon:
* Check language support on the search

wcandillon:
* Check support for fuzzy search

PR #709, opened 2015-02-16
Fix promise resolving

wcandillon:
@dknochen is this correct?


tasks/s3.js
```diff
@@ -138,7 +138,7 @@ var createBucket = function() {
             defered.reject();
         } else {
             $.util.log('createBucket(' + bucketName + ')');
-            defered.resolve();
+            waitForBucketExists().then(defered.resolve);
```
dknochen: looks correct to me. Now, the waitforBucketExits is called twice: https://github.com/28msec/cellstore/pull/709/files#diff-c8389580628a68556a620974b4694c2eR200

wcandillon: @dknochen I don't see it. Where?

wcandillon:
@dknochen This was a big mixup. master wasn't merge in my branch and I was looking at an outdated version of the code.

PR #714, opened 2015-02-17
Add default default value if requesting all values of a dimensi...

gislenius:
...on.


wcandillon:
@gislenius it works great!

PR #721, opened 2015-02-18
Restxq response format

fcavalieri:


gislenius:
Hi Federico,

It looks awesome! When do you plan to do the audit workaround?

fcavalieri:
Hi Ghislain, thanks.  I am currently working on the next release of 28th
that is a prerequisite for the branch.  I should get to the workaround
today.
Best
Federico

wcandillon:
@fcavalieri If `GET`s are enabled now, we need to update the swagger file.

fcavalieri:
I agree.  This is still not the case.
Best
Federico

gislenius:
@fcavalieri I think we should take the opportunity of this PR to make these endpoints all non-sequential/GET (that is, making audit non-deterministic). The performance gain would be significant, because a lot of clauses would no longer get materialized in vain.

fcavalieri:
I agree. This is my goal as well and the reason yesterday Sausalito was
updated to 4.0.1 :)

fcavalieri:
@gislenius The workaround I was talking about cannot work in our current settings. (see William's email, or chat on skype)
The workaround cannot work on the mongo we use in production for the portal deployment.
Thee ability of inserting documents through "run-command" has been introduced in 2.6 and the portal deployment is 2.4
so I see three alternatives:
 1) dropping the audit (on which I have no opinion as I don't know its reasons of existence)
 2) Upgrading the portal deployment to 2.6
 3) introducing an undocumented non-sequential mongo:insert

fcavalieri:
Hello @wcandillon , can you re-review this when you have time? Thanks

gislenius:
Hi Federico,

I'd go for dropping the audit.

The only effect I can see is that users won't have any API call counter any
more, but since we changed the pricing model it's useless now.

Kind regards,
Ghislain

fcavalieri:
Hi Ghislain,
thanks for the feedback.
The PR removes the audit calls and makes all api/* queries non-sequential.
Best
Federico

gislenius:
If William approves, I approve too! ;-)

wcandillon:
This is awesome!

PR #724, opened 2015-02-19
Eidaid fix

gislenius:


queries/public/api/entities.jq
```diff
@@ -49,7 +49,8 @@ let $entities := multiplexer:entities(
 let $entities :=
   for $entity in $entities
   return {|
-    project($entity, "_id")[not $profile-name eq "japan"],
+    project($entity, "_id")[$profile-name eq "sec"],
+    { EID: $entity._id }[empty($entity.EIDs)],
```
dknochen: are we not returning the EIDs for the japan profile?

queries/public/api/entities.jq
```diff
@@ -49,7 +49,8 @@ let $entities := multiplexer:entities(
 let $entities :=
   for $entity in $entities
   return {|
-    project($entity, "_id")[not $profile-name eq "japan"],
+    project($entity, "_id")[$profile-name eq "sec"],
```
dknochen: is this kept for backwards compatibility?

queries/public/api/filings.jq
```diff
@@ -86,8 +86,8 @@ let $summaries :=
   default return
     for $archive in $summaries
     return {|
-      project($archive, "_id"),
```
dknochen: here its not kept for backwards compatibility.

queries/public/api/entities.jq
gislenius: Yup.

queries/public/api/entities.jq
gislenius: In the Japanese profile, there is an EIDs field containing the EIDs and _id contains a UUID that should be hidden.

queries/public/api/filings.jq
gislenius: We only ensure backwards compatibility for the SEC at the moment.

queries/public/api/entities.jq
dknochen: ok, found it:
trim($entity, "_id")

PR #725, opened 2015-02-19
Add archive statistics to output.

gislenius:


gislenius:
http://edinet-dev.28.io/v1/_queries/public/api/filings.jq?_method=POST&token=foobar&eid=http%3A%2F%2Fdisclosure.edinet-fsa.go.jp%20E01264-000&eid=http%3A%2F%2Finfo.edinet-fsa.go.jp%20E01264-000&eid=http%3A%2F%2Fwww.tse.or.jp%2Fsicc%2054110&format=html&profile-name=japan&fiscalYear=ALL&fiscalPeriod=ALL

queries/public/api/filings.jq
```diff
@@ -95,8 +95,20 @@ let $summaries :=
               format: $format,
               profile-name: $profile-name
           })
-      },
-      trim($archive, "_id")
+       },
+      trim($archive, ("_id", "Statistics")),
```
wcandillon: @gislenius Can we get rid the of the `Statistics ` object key all together?

gislenius: I don't understand what you mean here. The Statistics key is in the archive object stored in MongoDB. The call to get the number of each report elements use these statistics.
What we could do in the future though, is to statically add the Num* fields already in the infoset generator, and potentially remove the Statistics part. But I would postpone this until later to not delay the next infoset generation.

dknochen:
@gislenius Is the idea of this branch to inline the Statistics? To make it more table like? 
The link you posted doesn't contain any of the changes, I think.

queries/public/api/filings.jq
```diff
@@ -95,8 +95,20 @@ let $summaries :=
...
+          NumFacts: archives:num-facts($archive),
+          NumFootnotes: archives:num-footnotes($archive),
+          NumReportElements: archives:num-report-elements($archive),
+          NumHypercubes: archives:num-hypercubes($archive),
+          NumDimensions: archives:num-dimensions($archive),
+          NumMembers: archives:num-members($archive),
+          NumLineItems: archives:num-line-items($archive),
+          NumAbstracts: archives:num-abstracts($archive),
+          NumConcepts: archives:num-concepts($archive)
+      }
```
dknochen: why not just $archive.Statistics?

gislenius: Because the Statistics object is very messy and hierarchical (plenty of useless numbers that will later disappear) and not supposed to be exposed to users.

gislenius:
The link was obsolete because another deployment was done. I'm redeploying to a specific backend.

gislenius:
@dknochen The link now works.

queries/public/api/filings.jq
dknochen: should we then not do the same cleanup for the SEC profile? Backwards compatibility for the Statistics field doesn't seem that important. What do you think?

dknochen:
thanks the link worked, very nice :)

image

gislenius:
@dknochen done ;-)

gislenius:
Back to you!

queries/public/api/filings.jq
gislenius: DONE

PR #728, opened 2015-02-20
retry putting website if bucket does not exist.

dknochen:


wcandillon:
@dknochen Can you explain me this change?

dknochen:
although, we wait for the bucket to exist before we upload to it, we somtimes get a NoSuchBucket exception when we start the upload. I cannot explain why - its just eventual consistency. For this case, I simply added a retry because the bucket does exist.

image
https://circleci.com/gh/28msec/cellstore/194

wcandillon:
@dknochen Can you investigate what `waitFor` does exactly?

wcandillon:
@dknochen I think it need to really investigate what's going on here.

dknochen:
s3.waitFor('bucketExists', params = {}, [callback]) ⇒ AWS.Request

"Waits for the bucketExists state by periodically calling the underlying S3.headBucket() operation every 5 seconds (at most 20 times)."

http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#bucketExists-waiter

wcandillon:
@dknochen I'm not confortable merging this at the moment. Let's see at which rate this happen. What I would be comfortable with  is to call `createBucket` twice everything (the second one being idempotent).
Do you agree?

dknochen:
@wcandillon Your suggestion doesn't make sense to me. Creating the bucket is not the problem. It is created correctly. It's only about when it becomes visible.
Can you provide a PR?

wcandillon:
@dknochen Let's discuss it offline.

PR #729, opened 2015-02-20
Update to japanxbrl-8_6

wcandillon:


gislenius:
Ouch. That should not have been merged until we do manual checks.

PR #733, opened 2015-02-23
Add test for inter-gen hypercube.

gislenius:


queries/public/test/edinet/facts.jq
```diff
@@ -36,5 +36,18 @@ test:check-all-success({
           },
           function($res as item*) as item* { count($res.FactTable[]) },
           22
+        ),
+    inter-generation: test:invoke-and-assert-deep-equal(
+          "facts",
+          {
+              "edinetcode" : "E04147",
+              "fiscalPeriod": "FY",
+              "fiscalYear": "ALL",
```
dknochen: with ALL it will fail in the future when a new filing for the entity arrives. Could we explicitly put the years that we have so far?

dknochen:
:+1:

queries/public/test/edinet/facts.jq
gislenius: DONE

PR #738, opened 2015-02-23
Test ambiguous concepts bug

dknochen:


queries/modules/io/28/apps/test.jq
```diff
@@ -12,6 +12,10 @@ declare function test:is-dow30() as boolean{
   contains(credentials:credentials("MongoDB", "xbrl").db, "dow30")
 };
 
+declare function test:is-all() as boolean{
```
gislenius: I would rename this to make clear it's japan.

queries/public/test/edinet/facttable-for-component.jq
```diff
@@ -0,0 +1,17 @@
...
+        ambiguous-concepts-fix: test:invoke-and-assert-deep-equal(
+          "facttable-for-component",
+          {
+            aid: "TD2014110700368",
+            role: "http://disclosure.edinet-fsa.go.jp/role/jppfs/rol_QuarterlyConsolidatedStatementOfCashFlows-indirect"
+          },
+          function($res as item*) as item* { count($res.FactTable[]) },
+          78
+        )
+    }[$is-all]
```
gislenius: Can you also add a test for a NIKKEI-9 filing? The bug we fixed was in all BS/IS/CS in TDNET filings.

gislenius:
@dknochen back to you! ;-)

queries/public/test/edinet/facttable-for-component.jq
dknochen: @gislenius I couldn't find any change due to this change in any top-9 filing. Do you have a specific example at hand? Or, was it only a guess?

PR #748, opened 2015-02-26
Write more documentation.

gislenius:


wcandillon:
Introduction 
* I find the first paragraph a bit awkward.
* I would remove second paragraph.
* The structure after the Cell stores section needs to be reworked (to be discussed offline).
* Link to the API documentation (as an annex?)

Entities
* FSA is not defined
* Profile is not defined
* Is there a glossary?
* Show a entities listing before selecting a single one (self containment).
* The details about the output formats are repeated across chapters but not consistently.

Archives
* Archives can be though of Excel files => though of as?
* Show an archive listing before selecting a single one (self containment).
* Why is the available periods API endpoint section there? I can give suggestions about this offline.

Componnents
* It would be nice to drag the excel metaphor across all chapters and then give the full metaphor in the introduction.

Many chapter contain list definitions, check gitbook documentation if there is a better formating for this.

wcandillon:
Overall it's very very good. It really shows the power and the simplicity of the technology. Also I really love some of the structure.
Some comments:

Facts
* There are some formatting problems with the JSON.
* Fetching Facts: We might want to use some tabular representation, or maybe some sort of figure.
* Default Dimension Values: Is it possible to give a concrete example?
* Is the concept of label introduced?
* Fact table: make a concrete example query and results. I felt that the first chapter might be too vague for someone not famillar with the matter.

Model Structure
* Link to previous chapter?
* The hierarchy example might be a tiny bit too long

gislenius:
Thanks William. The comments on the introduction and on entities have been addressed.

wcandillon:
@gislenius This is how glossary seem to be handled by gitbook: https://github.com/GitbookIO/gitbook#glossary

gislenius:
@wcandillon Thanks! How does the plugin work though?

wcandillon:
Can you double check the casing of the titles? They seem inconsistent overall.

* 0.1 Current Technologies => I would absolutely remove it.
Especially in the context that cell stores are targeted to business users.
I would replace it with a section about XBRL. This would help make the relationship between a cell store and XBRL more explicit. A reader might not know XBRL.

* 0.2 Cell Stores => "Think of taking an Excel spreadsheet and a pair of scissors".
It is not explicitly explained what putting a trillion cells in a bad does for the business user. 
"I mean, seriously, how many of us have tried to unzip and open an Excel file with a text editor for any other reason than mere curiosity?"
=> Is there a way to rephrase in a way that make the comparison more explicit? Something like: "XBRL is to cell store what the inside format of .xsls files are to Excel. Seriously, ...."

* 0.4 Analogy With spreadsheets. Is there are problem with the casing of the title? The spreadsheet column should be on the left. I cannot make sense of the first figure. I am not sure I understood the last section.

* 0.5 The REST API. 
"All queries shown in this tutorial can be run directly on our platform." => The jump is too abrupt. We need to explain in one setence or two that the cellstore provides a REST API.
"Private usage of this token is free and we only bill for commercial use." => "Private usage of this token is free. For commercial use, contact us at...".
Should you give the token?

2. Archives. "What is inside (i.e., the XBRL syntax) is irrelevant to business users. These details are hidden to them." => I would either remove this setence or make the point you are trying to make more explicit.

3. Components. Do component contain both metadata and data? If they only contain metadata shouldn't the spreadsheet analogy be that they are excel templates? And report elements, elements of an excel template?

wcandillon:
My part:
* Bidirectional links between the REST API and this documentation.
* Runnable examples.

gislenius:
- 0.1 Current Technologies => I would absolutely remove it.

This documentation is intended for developers, not for business users.

gislenius:
- I would replace it with a section about XBRL.

I think a section on XBRL totally makes sense.

gislenius:
- It is not explicitly explained what putting a trillion cells in a bad does for the business user. 

This is not intended to be read by business users. I can write more on scalability, but again this is intended for developers.

gislenius:
- => Is there a way to rephrase in a way that make the comparison more explicit? Something like: "XBRL is to cell store what the inside format of .xsls files are to Excel. Seriously, ...."

agreed.

gislenius:
- 0.4 Analogy With spreadsheets. Is there are problem with the casing of the title?

I went through them all to make consistent, that is, all words in H1 titles are capitalized, and only the first letter is capitalized in H2+ titles. I probably missed some.

gislenius:
- The jump is too abrupt. We need to explain in one setence or two that the cellstore provides a REST API.

Agreed.

gislenius:
- Should you give the token?

I will.

gislenius:
- I would either remove this setence or make the point you are trying to make more explicit.

I'll make it more explicit.

gislenius:
- Do component contain both metadata and data

From a user/UI perspective, they contain both. They are really like a sheet, not a template. It's just that components may share facts, so it's decoupled.

From an infoset perspective, component objects only contain metadata, but this metadata is kind of a "link" to the data it contains.

gislenius:
- And report elements, elements of an excel template?

Report elements really correspond to header (non-data) cells.

gislenius:
@wcandillon Back to you!

gislenius:
Thanks for your useful comments.

wcandillon:
@gislenius Excellent!

PR #757, opened 2015-02-27
Two fixes.

gislenius:


dknochen:
@gislenius the fix looks good to me :+1: . Could you add a test that used to fail before this fix? Thank you.

gislenius:
@dknochen Writing tests is not straightforward, as nothing in the secxbrl or edinet would be an example of fails-to-succeeds behavior. These errors were fixed against the early Swiss taxonomy. The first error prevents a crash if some labels are set to null (which normally should not happen if it comes from the infoset generator). The second error prevents an empty spreadsheet from appearing as broken, by forcing at least an empty column.

dknochen:
@gislenius Maybe, we have to generate swissxbrl-6 with dummy labels then instead of nulls?

gislenius:
Hi Dennis, that's a good point. I would also be fine with this.

PR #759, opened 2015-03-02
Change default value of fiscal year and period to the empty seq...

gislenius:
...uence.


gislenius:
@dknochen You can review ;-)

gislenius:
(this is as discussed previously)

dknochen:
it looks good. Just needs to be carefully documented in swagger/queries.json

gislenius:
@dknochen back to you ;-)

queries/public/api/facts.jq
```diff
@@ -212,9 +212,9 @@ declare  %rest:case-insensitive %rest:distinct  variable $edinetcode        as s
 declare  %rest:case-insensitive %rest:distinct  variable $tag               as string* external;
 declare  %rest:case-insensitive %rest:distinct  variable $ticker            as string* external;
 declare  %rest:case-insensitive %rest:distinct  variable $sic               as string* external;
-declare  %rest:case-insensitive %rest:distinct  variable $fiscalYear        as string* external := "LATEST";
-declare  %rest:case-insensitive %rest:distinct  variable $fiscalPeriod      as string* external := "FY";
-declare  %rest:case-insensitive %rest:distinct  variable $fiscalPeriodType  as string* external := ("instant", "YTD");
```
dknochen: this means we will also return QTD by default. You sure you wanted this?

dknochen: The documentation wasn't updated for it.

gislenius: It's intented to relax fiscalPeriodType filtering.

However, we could still discuss whether to still automatically filter YTD+instance in case the user does filter for a fiscal period or year.

I don't feel strong about it, even though I think it would be simpler to just have as little as possible of this "automatic behavior".

gislenius: I fixed the documentation.

PR #762, opened 2015-03-02
Fix build

wcandillon:


wcandillon:
@dknochen I'm trying to fix the regression. Is this PR correct? Or should it be `err.getMessage().indexOf('NoSuchBucket:') === 0` ?

dknochen:
I'm not sure about this. In the documentation it says its an error object: http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#putBucketWebsite-property
Can you confirm this?

wcandillon:
if you look at the latest master failure on circleci you will see the issue.

dknochen:
then yes, should be 

    _.isString(err.message) && err.message.indexOf ...

http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Response.html

wcandillon:
@dknochen back to you.

wcandillon:
@dknochen I think it still doesn't work.

PR #763, opened 2015-03-03
Config fixes

dknochen:
- several fixes for private cellstore deployment configurations
- allow using proxies

depends on https://github.com/28msec/28/pull/63


dknochen:
@wcandillon this is now ready for review. Thank you.

tasks/tests.js
```diff
@@ -22,6 +23,9 @@ gulp.task('tests:e2e', ['tests:update-webdriver'], function() {
   }
   var specs = Config.paths.e2eSpecs;
   $.util.log('specs: ' + specs);
+  if(!_.isArray(specs) || specs.length === 0){
```
wcandillon: @dknochen Why this change?

wcandillon:
@dknochen I am not sure why you authorize empty spec lists, other than that it's good.

tasks/tests.js
dknochen: Otherwise, you have to create an e2e test before creating and testing a new cellstore config.

dknochen:
Otherwise, you have to create an e2e test before creating and testing a new cellstore config. Its just to remove the obstacles. Alternatively, we could add a default e2e test that is run for any deployment (e.g. the auth-scenario.js). What do you think?

wcandillon:
@dknochen I like the idea to have the e2e auth test running for all scenarios.

dknochen:
@wcandillon I also like it :)

PR #765, opened 2015-03-05
Add parameter for open hypercube queries.

gislenius:


gislenius:
@wcandillon this should simplify everybody's life when querying facts not knowing about dimensions in advance. open=true makes an open hypercube query, meaning it does not reject facts with more dimensions.

swagger/queries.json
```diff
@@ -2365,6 +2371,18 @@
                 }
             },
             {
+                "title": "Get all dividend forecasts for 2014 reported by Canon. Open hypercube semantics.",
```
dknochen: its not only forecasts. The result will contain results and forecasts including all segmenting dimensions.

gislenius:
@dknochen back to you ;-)

PR #767, opened 2015-03-05
Add guard for unit fixing to restrict to SEC.

gislenius:


gislenius:
This fixes an inconsistency (see facts.jq). In the general case, units should not be handled in any special way and are just another aspect.

PR #771, opened 2015-03-05
Fix non us testing

dknochen:


dknochen:
@wcandillon could you review this?

PR #777, opened 2015-03-06
Migrate to secxbrl-8_8.

gislenius:


gislenius:
The only remaining error is the name of AT&T that got encoded by mistake. As soon as this is fixed, we'll repopulate all-sec-filings-8.

fcavalieri:
I fixed the error. I will merge this as soon as the tests finish

PR #781, opened 2015-03-09
fix check NoSuchBucket error

dknochen:


dknochen:
image

dknochen:
@wcandillon we could also use the retryDelay field, but maybe thats overdoing it?

wcandillon:
yay!

wcandillon:
@dknochen What's `retryDelay`?

dknochen:
@wcandillon in the screenshot above you can see that "retryDelay: 30" is part of the response of S3. I think they want to suggest us that we should wait for 30 Seconds until we try again. Anyways, I don't believe we should follow this.

wcandillon:
Thks for the info! I agree.

PR #791, opened 2015-03-11
Remove config.json file

wcandillon:


wcandillon:
@dknochen an good honest refactoring.

tasks/templates.js
```diff
@@ -8,40 +8,23 @@ var $ = require('gulp-load-plugins')();
 var _ = require('lodash');
 var mkdirp = require('mkdirp');
 var path = require('path');
+var Mustache = require('mustache');
 
 var Config = require('./config');
 
-var loadCredentials = function(){
+gulp.task('templates:config', [ 'crypt:decrypt' ], function(){
```
dknochen: I think, there is no reason any more why this task should stay in templates.js. 
We should move this rest into config.js as well.

dknochen:
@wcandillon please see: https://github.com/28msec/cellstore/pull/794

wcandillon:
@dknochen Nice catch! That actually optimizes a lot.

wcandillon:
@dknochen back to you

PR #794, opened 2015-03-12
No mustache

dknochen:


wcandillon:
@dknochen Looks good. Can you give me more info on the motivation for this? Why is it done on these file specifically and not everywhere?

dknochen:
@wcandillon I need this change to be able to define multiple datasources in the encrypted config. We need to add an elastic search data source. With mustache it simply doesn't work.

wcandillon:
@dknochen I would be happy to review and merge this but the branch is conflicting.

dknochen:
superseded: https://github.com/28msec/cellstore/pull/800

PR #797, opened 2015-03-12
Minor renaming

wcandillon:


tasks/lint.js
```diff
@@ -32,4 +32,4 @@ gulp.task('lint:xqlint', function(){
         .pipe($.xqlint.failReporter());
 });
 
-gulp.task('lint:run', ['lint:jslint', 'lint:jsonlint', 'lint:xqlint']);
+gulp.task('lint', ['lint:jslint', 'lint:jsonlint', 'lint:xqlint']);
```
dknochen: this should be moved in the main gulpfile.js then. The naming convention is that if its in a tasks file it has a prefix like this: "<file>:<function>".

wcandillon:
Make it consistent.

wcandillon:
@dknochen I did some cleanup. Let me know what you think.

dknochen:
:+1:

PR #799, opened 2015-03-12
Fix XQLint memory leak

wcandillon:


tasks/lint.js
```diff
@@ -26,10 +29,39 @@ gulp.task('lint:jsonlint', function(){
...
+
+    var spaces = function(count){
+        var result = '';
+        for(var i = 1; i <= count; i++){
+            result += ' ';
+        }
+        return result;
+    };
+
+    var files = expand(Config.paths.jsoniq[0]).concat(expand(Config.paths.jsoniq[1]));
```
dknochen: better use http://underscorejs.org/#reduce for this

wcandillon: thank you! I will fix it.

PR #808, opened 2015-03-14
Minor refactoring

wcandillon:


tasks/lint.js
```diff
@@ -31,14 +31,6 @@ gulp.task('lint:jsonlint', function(){
...
-        var result = '';
-        for(var i = 1; i <= count; i++){
-            result += ' ';
-        }
-        return result;
-    };
-
     var files = expand(Config.paths.jsoniq[0]).concat(expand(Config.paths.jsoniq[1]));
 
     files.forEach(function(file){
```
wcandillon: @dknochen You suggested to use reduce but I am not 100% where. Instead of the foreach? To accumulate the markers?

wcandillon:
@dknochen You suggested to use reduce but I am not 100% where. Instead of the foreach? To accumulate the markers?

dknochen:
@wcandillon I meant something like this: https://github.com/28msec/cellstore/pull/811

wcandillon:
Make sense! :+1: . The identation seems wrong (weird that jshint doesn't catch it). I re-started the build without caching that should make it pass (because of lodash 3.5).

PR #811, opened 2015-03-16
use reduce instead of hardcoded array.

dknochen:


wcandillon:
Make sense! :+1: . The identation seems wrong (weird that jshint doesn't catch it). I re-started the build without caching that should make it pass (because of lodash 3.5).

PR #816, opened 2015-03-17
Improve csv/html display of nested arrays.

gislenius:


dknochen:
there is a full stop AND a comma:
image

PR #824, opened 2015-03-18
Move new report frame to new location.

gislenius:


gislenius:
There was a concurrent change. The reports directory was changed, but the new report frame was still added to the old location.

This PR fixes this.

PR #828, opened 2015-03-18
Improve elastic search

dknochen:


queries/modules/io/28/modules/xbrl/profiles/japan/core.jq
```diff
@@ -35,6 +35,8 @@ declare %an:nondeterministic function japan:entities(
     $eid as string*,
     $tag as string*,
     $full-text as string?,
+    $full-text-offset as integer?,
```
gislenius: I would rather add an $options parameter, the usual way

dknochen:
@gislenius adressed your comments

wcandillon:
@gislenius do you approve correct?

PR #839, opened 2015-03-20
Dimensions tab

dknochen:


dknochen:
@wcandillon could you have a first glance?

wcandillon:
@dknochen I'm trying to figure out how the build passes :confused:

wcandillon:
@dknochen nevermind ;-)

app/report/dimensions/dimensions.html
```diff
@@ -0,0 +1,53 @@
+<div id="dimensions">
+    <div class="scrollable content">
+        <header class="std-header">Segmentation / Filtering by Dimensions</header>
```
wcandillon: Slice & Dice

app/report/report.html
```diff
@@ -9,6 +9,7 @@
         <div>
             <ul class="nav navbar-nav">
                 <li ui-sref-active2="active" ui-sref-active-includes="report.taxonomy"><a ui-sref2="report.taxonomy.concepts()" id="taxonomy-link">Taxonomy</a></li>
+                <li ui-sref-active2="active" ng-show="displayDimensionsPage"><a ui-sref2="report.dimensions()" id="dimensions-link">Segmentation / Filtering</a></li>
                 <li ui-sref-active2="active" ng-show="displayFiltersPage"><a ui-sref2="report.filters()" id="filters-link">Filters</a></li>
```
wcandillon: I don't like this switch, what's the reasoning behind it?

app/report/report.js
```diff
@@ -3,11 +3,8 @@
 angular.module('report-editor')
 .controller('ReportCtrl', function($q, $rootScope, $scope, $log, PROFILE, Session, API, Report, report){
     $scope.report = report;
-    $scope.displayFiltersPage = false;
-
-    if(PROFILE === 'sec') {
-        $scope.displayFiltersPage = true;
-    }
+    $scope.displayFiltersPage = (PROFILE === 'sec');
```
wcandillon: :(

Should we take advantage of this PR to remove the Too generic feature both backend and in a frontend. Maybe it could be done in another PR.

wcandillon:
I like it.

I have some other questions:
* Does it make sense to have the filter on concepts? Should this be handled in the taxonomy? Or maybe at least not show by default in the UI. For now I guess I'd be happy if it just show on the bottom or something.
* Can the default filter for xbrl:Entity be updated to ALL instead in the Japanese FAC example?

wcandillon:
@dknochen my comment on the concept filter was a bit strange and confused. But still there might some room for improvement there.

wcandillon:
Probably for another PR but the top menu doesn't make sense anymore. Taxonomy, Slice & Dice, and Visualize are steps. The UI could make this more explicit.

gislenius:
@dknochen looks awesome! Some comments:

1. I agree with William that concepts filtering is already handled by the taxonomy editor and should maybe not be in this tab, which is specific to other aspects.
2. I think the second tab name should be changed to "Dimensions", or "Slice" (it doesn't contain Dice yet).
3. In the list of members for each dimension, the logical "or" is missing to make clear it is "equals or equals or equals" etc.
4. When I edit dimensions, like add one, and then go to facts, it doesn't seem to be saved at all. When I go back, it's gone.

I hope it helps.

gislenius:
5. If you remove the period dimension and add it back, there seems to be errors.

dknochen:
@wcandillon @gislenius its ready for review. It is a pretty big PR. I could split it into smaller chunks if you prefer ... Let me know...

queries/public/export/edinet/fac.jq
```diff
@@ -1,39 +1,126 @@
...
 replace value of json $report.Hypercubes."xbrl28:ImpliedTable".Aspects."xbrl:Period" with
     {
       "Name": "xbrl:Period",
       "Label": "Period [Axis]",
       "Kind": "TypedDimension",
-      "Type": "string"
+      "Type": "string",
+      "Id" : "8ca7b4c4-a2a2-4897-ab5a-a4f8472e097a"
+    };
+replace value of json $report.Hypercubes."xbrl28:ImpliedTable".Aspects."xbrl:Unit" with
```
wcandillon: @dknochen I'm not clear on what this does.

app/modules/uuid.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+angular.module('uuid', [])
+.factory('UUID', function() {
```
wcandillon: @dknochen Can you use this in other places where uuids are used?
Also we might want to remove `AbstractReport.prototype.uuid` then.

app/modules/models/table-sec.js
```diff
@@ -0,0 +1,156 @@
+'use strict';
+
+angular
```
wcandillon: @dknochen Let's discuss tables offline.

app/modules/models/member-explicit.js
```diff
@@ -0,0 +1,75 @@
+'use strict';
```
wcandillon: @dknochen Not sure what this does yet.

wcandillon:
@dknochen The code is very well written. I have some questions.
Let's discuss it tomorrow in the team meeting.

gislenius:
1. I can't add xbrl:Scenario as a dimension, because as soon as I type xbrl: it gets grayed and I can't remove it
2. When you add a default value, I think it should be also added on the right to the filters, but grayed. Otherwise you need to always add it manually, twice.
3. When no filter is in place, it could make sense to make it more visible that no filter is in place
4. I think the standard aspects should come first on the page. Perhaps some text could be added that says "This builtin dimension cannot be removed."

gislenius:
We could add some explanatory text.

In each standard aspect, I would write "This standard dimension cannot be removed."

In each other aspect, I would write either "This dimension has no default value. Only facts that associate it with one of the filters on the right will be output." or "This dimension has a default value. Facts that associate it with one of filters on the right will be output. Facts that do not associate it with any value will also be output, with an association to the default value." depending on whether a default value is specified.

On top of the page, I would add "

This page allows you to slice and dice hypercubes. Hypercubes are made of dimensions. You can add or remove dimensions, and specify, for each dimension, which values are included.

For a fact to be included in this hypercube, it must (i) contain all the dimensions on this page , implicitly or explicitly, and (ii) satisfy all dimension filters, that is, associate each dimension on this page with one of the values (members) listed in the filters for that dimension.

Default values allow you to include a fact even if it does not contain one of the dimensions. It will then be implicitly associating this dimension to its default value."

app/app.js
```diff
@@ -12,6 +12,8 @@ angular.module('report-editor', [
     'api',
     'session-model',
     'report-model',
+    'table-model',
```
wcandillon: to rename

wcandillon:
@dknochen Isn't it what we need to clean up the models: http://www.typescriptlang.org/Tutorial?

PR #840, opened 2015-03-23
Count

gislenius:
Customer request: perform a count only, without downloading the actual data (faster performance).


gislenius:
http://edinet-dev.28.io/v1/_queries/public/api/facts.jq?token=foobar&ticker=7751&fiscalYear=2014&fiscalPeriod=FY&count

dknochen:
the actual user request could be added as example and sec test.

dknochen:
Very clean PR. I like that :)!

gislenius:
@dknochen the use case has been added to both the documentation and tests. It will be faster once we have the entity index on the facts collection.

PR #848, opened 2015-03-24
fix deep equal array comparison

dknochen:


queries/modules/io/28/apps/test.jq
```diff
@@ -223,7 +223,7 @@ declare function test:deep-equal-no-array-order(
     if(count($expected) ne count($actual))
     then false
     else
-      every $i in count($expected)
+      every $i in (1 to count($expected))
```
gislenius: You don't need the extra parentheses.

gislenius:
Hi Dennis,

That's an awesome catch! Just a minor, cosmetic comment.

Kind regards,
Ghislain

dknochen:
@gislenius back to you

PR #850, opened 2015-03-25
two cellstore bug fixes

dknochen:


app/report/report.html
```diff
@@ -2,7 +2,7 @@
     <header class="navbar navbar-default">
         <div class="navbar-header">
             <span class="navbar-brand">
-                <a class="fa fa-list" href="../" id="reports"></a>
+                <a class="fa fa-list" href="/" id="reports"></a>
```
wcandillon: :+1:

app/report/taxonomy/concept/concept.html
```diff
@@ -23,7 +23,7 @@
         </li>
         <li ng-if="!isValidationConcept" ui-sref-active2="active"
             ng-class="{ disabled: (concept.IsAbstract === true || isValidationConcept === true)}">
-            <a ui-sref2="report.taxonomy.concept.formula.validations.list()">Validations ({{report.listValidatingRules(concept.Name).length}})</a>
+            <a ui-sref2="report.taxonomy.concept.formula.validations.list()" id="validations-link">Validations ({{report.listValidatingRules(concept.Name).length}})</a>
```
wcandillon: Can you use class instead?

app/modules/report-model.js
```diff
@@ -1451,10 +1451,10 @@ angular
 
         var result = [];
         var model = this.getModel();
-        if (!_.isObject(model) || _.isArray(model.Rules)) {
```
wcandillon: @dknochen Can you explain this change?

dknochen: it should only return an empty array if there is NO Rules array.

app/report/taxonomy/concept/concept.html
dknochen: we have put ids on the other ones, e.g. https://github.com/28msec/cellstore/blob/dennis-fixes/app/report/taxonomy/concept/concept.html#L12
Its easier to use in the e2e tests. What would be the benefit of using class instead?

wcandillon:
@dknochen Orthogonal to this PR but the report model is 1700 lines of codes.

app/report/taxonomy/concept/concept.html
wcandillon: Yes that's true but I kind of hate that. From my experience, ids make css much much harder and the code less modular. How would the selector in the test look like if we use class instead? I forgot why it makes the tests harder.

dknochen: it wasn't that hard actually:

element(by.css('a.validations-link'))

PR #851, opened 2015-03-25
Spreadsheet html

gislenius:


queries/modules/io/28/apps/test.jq
```diff
@@ -98,13 +98,15 @@ declare %an:nondeterministic function test:invoke-and-assert-deep-equal(
...
-declare %an:nondeterministic function test:invoke-and-assert-status(
+declare %an:sequential function test:invoke-and-assert-status(
   $endpoint as string,
   $parameters as object,
   $expected-status as integer
 ) as item
 {
-  let $actual-status := test:invoke($endpoint, $parameters)[1]
+  let $url as string:= test:url($endpoint, $parameters, true)
+  let $response as object := http-client:send-request({href: $url, method: "GET", "options" : { "follow-redirect" : false } })
```
dknochen: there is also an option "status-only": true which we could use. Now, that we use options anyways.

wcandillon:
@dknochen I let you review it.

PR #853, opened 2015-03-25
fix http connection errors.

dknochen:
@fcavalieri said that it might be an AWS issue.

image


dknochen:
@wcandillon what do you think about this?

wcandillon:
We do retry on both on-connection-error and on 503. Do we know what kind of AWS it can be? Is it possible that's only a circle ci thing?

dknochen:
@wcandillon a 503 would definitely be an AWS thing. The on-connection-error can have all sorts of reasons. We are addressing here mainly the on-connection-error. @fcavalieri can you confirm?

fcavalieri:
Yes, the on-connection-error might have different reasons, such as a network connectivity issue on the circleci machine (dns, routing, packet loss,...).

dknochen:
@wcandillon we could remove the 503?

wcandillon:
@fcavalieri @dknochen We should definitely remove the 503 here. My concern is that if that there are a lot of connection error appearing, we won't see it while it might affect our users. Do you guys share this concern as well?

fcavalieri:
Yes, it also concerns me. I might have an idea on the reason. I will look
into it after the call.

On Thu, Mar 26, 2015 at 4:27 PM, William Candillon <notifications@github.com

dknochen:
@fcavalieri any news on this?

fcavalieri:
@dknochen The HTTP connection issues I was experiencing were unrelated (and are now fixed). Is it possible that the connection issues you were having were due to the ddos on github? Do you still have connection issues? If so, how can I reproduce them?

wcandillon:
Since last Thursday, it happened only once for me. Which is more than acceptable.

fcavalieri:
FYI ddos on github is still in progress: https://status.github.com/

wcandillon:
@fcavalieri @dknochen Let's keep it on hold for now.

wcandillon:
@fcavalieri @dknochen feel free to reopen this one.

PR #856, opened 2015-03-26
Fix backend protocol

dknochen:


wcandillon:
Ok but why? Also the readme would need to be updated too right?

dknochen:
@wcandillon is is in the readme ;)

the reason is that we use http also for authentication which is a security issue. We fixed this already, but at some point it must have been reverted. So, its a regression really.

wcandillon:
:+1:

PR #859, opened 2015-03-26
redeploy swissxbrl

dknochen:


wcandillon:
@dknochen @gislenius do you guys want to add swiss XBRL to the patch-queue?

PR #861, opened 2015-03-26
New report frames.

gislenius:


gislenius:
@wcandillon could you look into the process out of memory problem? Thanks!

wcandillon:
I'm currently looking at it... more tomorrow.

wcandillon:
@gislenius UpdateReportSchema.jq is 22MB, maybe we over did this time? :)

wcandillon:
@dknochen Maybe we need to discuss this offline how we can fix this.

wcandillon:
@gislenius @dknochen In the meantime, should we by pass the xqlint check of this file? I think it's fair since it is a generated file. For javascript we never lint generated files.

dknochen:
@wcandillon I deacitvated it for now. Sorry, for the accidental commit to master :sob: ...

dknochen:
I also added an issue to improve: https://github.com/28msec/cellstore/issues/872

wcandillon:
@gislenius How files like COMID-BSC-CF1-ISM-IEMIX-OILY.json come from? How are they generated?

tasks/lint.js
```diff
@@ -31,9 +31,14 @@ gulp.task('lint:jsonlint', function(){
 
 gulp.task('lint:xqlint', function(done){
 
-    var files = _.reduce(Config.paths.jsoniq, function(result, path) {
+    var files = _.chain(Config.paths.jsoniq)
+        .reduce(function(result, path) {
                     return _.union(result, expand(path));
-                }, []);
+                }, [])
+        .filter(function(path){
```
wcandillon: Why is it using a different method than in line 21?

wcandillon: Yes that was a silly question, my apologies!

PR #882, opened 2015-03-30
Increase version.

gislenius:


dknochen:
I don't see an increased version in comparison to master? Can you double check? You need to rm the config before decrypt on master...

gislenius:
@dknochen I do see 7 when I decrypt. Or do you mean master already has 7?

gislenius:
You're right. Master's already on 7. Guess I didn't remove the file before decrypting.

PR #883, opened 2015-03-30
Always include references if they are available.

gislenius:


queries/modules/io/28/modules/xbrl/components.jq
```diff
@@ -1331,7 +1331,7 @@ declare function components:to-xbrl-annotation(
             xlink:type="simple"
             xlink:href="{$taxonomyName}_ref.xml"
             xlink:role="http://www.xbrl.org/2003/role/referenceLinkbaseRef"
-            xlink:arcrole="http://www.w3.org/1999/xlink/properties/linkbase" />[$options.include-references]
+            xlink:arcrole="http://www.w3.org/1999/xlink/properties/linkbase" />[exists($components.Concepts[].References[])]
```
dknochen: good improvement :) Should we create a function in the components module? For consistency?

gislenius:
Good point. Back to you @dknochen

dknochen:
approved, once it passes

PR #888, opened 2015-03-30
Fix master.

gislenius:


dknochen:
we should really start to add tests

PR #892, opened 2015-03-30
Allow filter override on component display.

gislenius:
#874


gislenius:
http://rendering.secxbrl.info/#?url=http:%2F%2Fsecxbrl-892.28.io:80%2Fv1%2F_queries%2Fpublic%2Fapi%2Fspreadsheet-for-component.jq%3Ftoken%3Dfoobar%26ticker%3Dko%26fiscalYear%3D2010%26fiscalYear%3D2011%26fiscalYear%3D2012%26fiscalYear%3D2013%26fiscalYear%3D2014%26fiscalPeriod%3DFY%26disclosure%3DBalanceSheet%26merge%3Dtrue%26override%3Dtrue%26format%3Djson

queries/public/api/facttable-for-component.jq
```diff
@@ -142,13 +155,7 @@ let $facts :=
...
             $component,
             {|
-                {
-                    FilterOverride : {
-                        "sec:FiscalPeriod" : { Type: "string", Default: null },
-                        "sec:FiscalYear" : { Type: "string", Default: null },
-                        "sec:FiscalPeriodType" : { Type: "string", Default: null }
-                    }
-                }[$profile-name eq "sec"],
+                { FilterOverride: $filter-override }[$profile-name eq "sec" and $override],
```
dknochen: why only for the sec profile here?

dknochen:
@gislenius I have 2 stupid questions: 
1. Why do we only allow an override for fiscal* Params? Could we also allow it for entities and aids? 
2. We have the same problem for reports. Could we add this param for reports as well? Or, is it a different issue?

thank you for this fix. Its beautiful.

gislenius:
Hi Dennis,

There is no such thing as a stupid question.

1. Why do we only allow an override for fiscal* Params? Could we also allow it for entities and aids? 

We could, however the use case is much less relevant. The reason is simple: that's because, when you ask to merge components, the facts *already* all correspond to the supplied entity and aids (simply because all facts in a filing match its aid as well as the submitting entity). Giving them overriding semantics won't change the output.

I'd generalize filter overrides separately, and would probably wait until we merge reports and components, because this is where it will be really super useful.

2. We have the same problem for reports. Could we add this param for reports as well? Or, is it a different issue?

We don't. The problem is simple: it's that the fiscal* parameters(and actually also the entity/aid parameters) have two uses across endpoints: filter components, filter facts. In *-for-component endpoints, they were used for component filtering. In *-for-report endpoints, they were used for fact filtering (=override).

This PR simply allows, in *-for-component endpoints, to also make them filter facts. In -for-report endpoints, that's already the case so there is nothing to change there.

I hope it makes sense!

dknochen:
@gislenius thanks for the extensive answers. It all makes sense to me. Yet, I think it only works for the sec profile so far, not for japan for example. This is why I thought it doesn't work for *-for-report endpoints.

Could you fix that it also works for japan? Probably by adding multiplexer::filter-override(....)? I can add an issue for the *-for-report endpoints (here also the eid and aid overrides should work, I think)...

Is what I'm seeing correct?

gislenius:
As discussed offline, we need to open an issue for adding this multiplexer function in another PR.

PR #895, opened 2015-03-31
Swiss XBRL fixes

gislenius:


dknochen:
@gislenius the fix is good. Should we start adding the first test?

gislenius:
@dknochen there is already a test that checks that it successfully returns.

queries/modules/io/28/modules/xbrl/concepts.jq
```diff
@@ -289,7 +289,7 @@ declare function concepts:tree-to-xbrl(
   let $children as object* := ($tree.To[], $tree.Members[])
   let $fromID as string := replace($tree.Name, ":", "_")
   let $fromLabel as string := $fromID || "_" || $level || "_" || $pos
-  let $fromLocator as element() := <link:loc xlink:type="locator" xlink:href="{$taxonomyName}.xsd#{$fromID}" xlink:label="{$fromLabel}" xlink:title="{normalize-space($tree.Label)}" />
+  let $fromLocator as element() := <link:loc xlink:type="locator" xlink:href="{$taxonomyName||"-"||tokenize($tree.Name, ":")[1]}.xsd#{$fromID}" xlink:label="{$fromLabel}" xlink:title="{normalize-space($tree.Label)}" />
```
dknochen: this won't work if the taxonomy is suffixed with "entry-point". Maybe, we should maintain an object that maps prefixes to taxonomy names and pass it along?

dknochen: Maybe, I'm confusing something here?

gislenius: I think that never happens -- the entry point has no concepts at all.

PR #916, opened 2015-04-07
remove maxcdn pull zone.

dknochen:
we don't have a pull zone for hq.secxbrl.info right now


dknochen:
@wcandillon can you review? We don't have a pull zone for hq.secxbrl.info right now. I removed it now, but we could also create a pull zone. What do you think?

wcandillon:
@dknochen no pull zones means no https right?

dknochen:
@wcandillon yes (only for the frontend, backend still has https)

PR #923, opened 2015-04-13
Now expanding hypercubes within networks.

### LINKED ISSUE #924: Fuzzy component vs. role semantics
In most cases, an extended link role has a single hypercube. However, in some cases, several hypercubes are grouped within the same extended link role. In such cases, a component corresponds to each of these hypercubes, not to the entire role.

At the time, it is wrongly implemented and a single component is showed for the entire role, with the hypercubes listed and picked at random upon querying.

gislenius:
Fixes #924 

This PR modifies the components.jq endpoint to fix the semantics of having several hypercubes in the same extended link role (in which case there is one component for each of these). In other words, if a role has several hypercubes, then one entry is created for each, as opposed to a single entry before.

Charles Hoffman's terminology:
Component = Network (or Extended Link Role) + Table (or Hypercube)

It also introduces a new hypercube parameter to filter on these (see documentation and test).

Two endpoints (spreadsheet, report-elements) will be upgraded with this new parameter in a subsequent PR, when the infoset is regenerated and the line items bug fixed.


gislenius:
http://secxbrl-tables.28.io/v1/_queries/public/api/components.jq?_method=POST&token=foobar&aid=0001214659-13-001731&format=html&role=http://www.brin.com/role/SupplementalCashFlowInformationDetails

http://secxbrl-tables.28.io/v1/_queries/public/api/components.jq?_method=POST&token=foobar&aid=0001214659-13-001731&format=html&role=http://www.brin.com/role/SupplementalCashFlowInformationDetails&hypercube=us-gaap:OtherSignificantNoncashTransactionsTable

http://secxbrl-tables.28.io/v1/_queries/public/api/components.jq?_method=POST&token=foobar&aid=0001214659-13-001731&format=html

queries/modules/io/28/modules/xbrl/components.jq
```diff
@@ -457,97 +462,155 @@ as object*
...
+declare function components:num-networks($component as object) as integer
 {
-  $components ! $$.Statistics.NumNetworks
+  $component.Statistics.NumNetworks
 };
 
 (:~
  : Return the number of report elements in each of the given components.
  :
- : @param $components sequence of components.
```
dknochen: in each of the given components -> in the given component

queries/modules/io/28/modules/xbrl/components.jq
```diff
@@ -1085,13 +1148,16 @@ declare %private function components:has-references(
  : @return all report elements without metadata.
```
dknochen: @param $hypercube an optional hypercube name.

dknochen:
These lines need to be adapted, too:

https://github.com/28msec/cellstore/blob/Tables/queries/public/api/facttable-for-component.jq#L144
https://github.com/28msec/cellstore/blob/Tables/queries/public/api/facttable-for-component.jq#L203
https://github.com/28msec/cellstore/blob/Tables/queries/public/api/facttable-for-component.jq#L216

dknochen:
:+1: very good :)

gislenius:
Back to you Dennis ;-)

I've also consolidated the num-* functions and added errors if hypercubes are not found.

queries/modules/io/28/modules/xbrl/components.jq
```diff
@@ -314,31 +324,40 @@ as object*
...
     $component as object,
+    $hypercube as string,
     $options as object?
 )
 as object*
 {
+  if(not $hypercube = (keys($component.Hypercubes), $components:ALL-HYPERCUBES))
+  then error(
+    QName("components:HYPERCUBE-DOES-NOT-EXIST"),
+    "Hypercube " || $hypercube || " was not found."
```
dknochen: if the hypercube does not exist in the component, an empty sequence cannot be used as string error will be raised, because the hypercube param is empty then: https://github.com/28msec/cellstore/blob/Tables/queries/public/api/facttable-for-component.jq#L123
Should we raise an error there instead?

dknochen: we could also factorize this part into the multiplexer and raise the error there:

image

dknochen:
I think if you address the final comment its good to be merged... :+1:

queries/modules/io/28/modules/xbrl/components.jq
gislenius: Done and done! ;-)

dknochen:
great stuff!! You just need to address the merge conflict which I couldn't do :|

PR #931, opened 2015-04-15
Implement hypercube switching for spreadsheet rendering

gislenius:
Dependency on #723 (merge it and back here before reviewing).


dknochen:
@gislenius I couldn't merge master into this one. There is a conflict. Could you take a look. Maybe, @fcavalieri can take a look tomorrow then?

gislenius:
Back to you @dknochen

queries/modules/io/28/modules/xbrl/components.jq
```diff
@@ -956,19 +1000,18 @@ declare function components:abstracts($components as object*) as string*
...
+    let $all-concepts := $component.Concepts[]
+    let $physical-line-items as object* := $all-concepts[$$.Kind eq "LineItems"]
+    return if(count($physical-line-items) le 1 and $hypercube ne "xbrl28:ImpliedTable")
            then $physical-line-items
            else
              let $presentation-network :=
                  networks:networks-for-components-and-short-names($component, "Presentation")
-             let $hypercube := descendant-objects($presentation-network)[$$.Kind eq "Hypercube"]
+             let $hypercube := descendant-objects($presentation-network)[$$.Name eq $hypercube]
              return $physical-line-items[$$.Name = $hypercube.To[].Name]
```
dknochen: what if lineitems are deeply structured? It will return only the 1st level.

queries/modules/io/28/modules/xbrl/profiles/sec/networks.jq
```diff
@@ -326,10 +327,18 @@ declare function sec-networks:standard-definition-models-for-components($compone
  : @error components:HYPERCUBE-DOES-NOT-EXIST if the specified hypercube is not found.
  : @return a definition model
  :)
-declare function sec-networks:standard-definition-models-for-components($components as object*, $options as object?) as object
+declare function sec-networks:standard-definition-models-for-components(
```
dknochen: should we rename the function if we only accept one component instead of components now?

queries/modules/io/28/modules/xbrl/components.jq
```diff
@@ -692,16 +709,21 @@ declare function components:standard-concept-breakdown(
...
  :
  : @param $component a component object.
+ : @param $hypercube a hypercube.
+ :
+ : @error components:HYPERCUBE-DOES-NOT-EXIST if the specified hypercube is not found.
  :
  : @return a definition model
  :)
-declare function components:standard-definition-models-for-components($components as object*) as object
+declare function components:standard-definition-models-for-components(
```
dknochen: should we rename the function if we only accept one component instead of components now?

queries/modules/io/28/modules/xbrl/components.jq
```diff
@@ -724,12 +747,24 @@ declare function components:standard-definition-models-for-components($component
  : </ul>
  :
  : @error components:HYPERCUBE-DOES-NOT-EXIST if the specified hypercube is not found.
+ :
  : @return a definition model
  :)
-declare function components:standard-definition-models-for-components($components as object*, $options as object?) as object
+declare function components:standard-definition-models-for-components(
```
dknochen: should we rename the function if we only accept one component instead of components now?

queries/modules/io/28/modules/xbrl/components.jq
gislenius: That cannot happen. LineItems are always children of hypercubes in presentation networks.

PR #910, opened 2015-04-02
Simplify and polish model structure.

gislenius:
Fixes #899 
Removes plenty of fields that are semantically irrelevant to model structures. Also added tabulations on the csv format for the report element labels to make structure explicit to human eye.

PR #933, opened 2015-04-15
Add tabs

gislenius:
Dependent on #910 to reduce diff.


dknochen:
As discussed yesterday, we should add a certain parameter which turns the tabs indentation on. By default there should be no tabs.
Do we agree on this?
thank you :)

gislenius:
... and there you go @dknochen  ;-)

gislenius:
Superseded by https://github.com/28msec/cellstore/pull/945

PR #937, opened 2015-04-16
Remove deprecated, unused functions.

gislenius:


gislenius:
... and there you go @dknochen ;-)

dknochen:
@gislenius did you test that the call to TotalForConceptGroupedByFiscalPeriodForAllEntities.jq still works? Adding a test for it would be too heavy weight, right?

gislenius:
@dknochen good catch, but it's already broken on master. I raised an issue.

https://github.com/28msec/cellstore/issues/953

I don't think it should block this PR.

dknochen:
@gislenius I agree. Thank you. :)

Issue #595, opened 2015-01-08
Deal with different swagger enums for different profiles


wcandillon:
How can we define different enums for different profiles in the swagger file? for example:

    {
        "name": "tag",
        "description": "A tag to filter",
        "profile": ["secxbrl"],
        "required": false,
        "type": "string",
        "enum": [
            "DOW30",
            "SP500",
            "RUSSEL1000",
            "FORTUNE100",
            "PJI",
            "ALL"
        ],
        "paramType": "query"
    }

and: 

    {
        "name": "tag",
        "description": "A tag to filter",
        "profile": ["edinet"],
        "required": false,
        "type": "string",
        "enum": [
            "NIKKEI",
            "ALL"
        ],
        "paramType": "query"
    }

wcandillon:
One option is to do:
```
{
  28-enums: {
     "sec": [
        "DOW30",
        "SP500",
        "RUSSEL1000",
        "FORTUNE100",
        "PJI",
        "ALL"
    ],
    "japan": [
        "NIKKEI",
        "ALL"
    ]
  }
}
```

PR #938, opened 2015-04-16
Add list of disclosures to documentation.

gislenius:


swagger/queries.json
```diff
@@ -691,6 +691,20 @@
                 "profile": ["sec", "japan"],
                 "required": false,
                 "type": "string",
+                "enum": [
```
dknochen: somehow we failed here because this list is only valid for the SEC. For japan we took different disclosure names:
  BalanceSheet
  StatementOfCashFlows
  StatementOfIncome

https://github.com/28msec/xbrl-infosetgenerator/blob/master/infosetgenerator/io/xbrl/profiles/fsa/FSAProfile.java#L175
maybe @wcandillon knows how we can have different enums for sec / japan?

gislenius:
That seems to be related to issue #595

PR #958, opened 2015-04-21
Disclaimer

gislenius:


wcandillon:
I like the idea and the result, don't like the dependency graph on this. Not sure what to think about it.
@dknochen any thoughts?

gislenius:
@wcandillon we could store this in a variable in the config module and provide is footer text in config.json...

wcandillon:
that'd be amazing.

gislenius:
@wcandillon can I hand it over to you?

dknochen:
to put it in the config file makes a lot of sense.
It should also be added to the other formats (not only html).
Not sure how we can add this to csv?

wcandillon:
I would use this formatting instead:
```html
              <small>Information is provided "as is" and solely for informational
                purposes, not for trading purposes or advice. secxbrl.info is provided as a complimentary REST API and aims at
                show-casing <a href="http://github.com/28msec/cellstore" target="_blank">28msec's cell store technology</a>. It delivers SEC data as it
                was submitted to the SEC and can be used for free, with no guarantee
                on its correctness or its availability whatsoever. By using it, you agree (i) to mention secxbrl.info as the source of
                the data, and (ii) that you consume this data at your own risk. In
                particular, this data does not constitute financial advice, and no
                investment decisions must be made based on this API. Please contact us
                if you wish your own cell store deployment with any data you like.
              </small>
```

PR #964, opened 2015-04-23
Parametrize datasources

wcandillon:


wcandillon:
@gislenius Does it achieve what you want?
@dknochen Is it clean?

tasks/templates.js
```diff
@@ -28,6 +29,14 @@ gulp.task('templates:create', ['config:load'], function(done){
         elasticSearchMapping = fs.readFileSync(mappingsFile, 'utf-8');
     }
 
+    //Indicates which datasource is for query and which one is for search
+    var queryDatasource  = _.chain(Config.credentials['28'].datasources)
+            .filter(function(source){ return source.type === 'query'; })
```
dknochen: if we make the type field an array, we could use the same datasource for query and search.

tasks/templates.js
```diff
@@ -28,6 +29,14 @@ gulp.task('templates:create', ['config:load'], function(done){
...
+            .map(function(source){ return _.template('{ name: "<%= name %>", category: "<%= category %>" }')(source); }).value() || '()';
+    var searchDatasource = _.chain(Config.credentials['28'].datasources)
+            .filter(function(source){ return source.type === 'search'; })
+            .map(function(source){ return _.template('{ name: "<%= name %>", category: "<%= category %>" }')(source); }).value() || '()';
```
dknochen: this will never be '()' because value() returns an empty array. We need to check for an empty array explicitly

dknochen: we should do  ....value()[0] || '()'

tasks/templates/config.jq.mustache
```diff
@@ -27,4 +27,8 @@ declare variable $config:sendmail-sender-name as string := "{{sendmail.sender.na
 declare variable $config:profile-name as string := "{{cellstore.profile}}";
 declare variable $config:filtered-aspects as integer := {{cellstore.filteredAspects}};
 
+declare variable $config:query-datasource := {{& queryDatasource }};
```
gislenius: can you add a type here?

gislenius: Technically, in the config file, I would need both the category and name of the source for both query and search. So are these objects with two fields?

tasks/templates.js
wcandillon: @dknochen can you elaborate?

tasks/templates.js
wcandillon: got it

wcandillon:
@dknochen Implemented the review comments

dknochen:
looks good :+1:

dknochen:
@gislenius approved from my side.

gislenius:
@dknochen from mine as well. Do you merge?

PR #965, opened 2015-04-23
Install Marklogic module automatically

wcandillon:


.gitignore
```diff
@@ -1,4 +1,6 @@
 node_modules
+jsoniq_modules
+queries/modules/io/28/modules/
```
gislenius: The XBRL modules are in here as well. Is it the intent? I would expect you to only want to ignore the marklogic module...

wcandillon:
@gislenius I fixed the entry but I really hope I find a more scalable solution for this in the future.

PR #969, opened 2015-04-23
Fix stats.

gislenius:


dknochen:
good catch :) ... 

there doesn't seem to be a test that checks the stats with several hypercubes, correct?

In order to keep you motivated, please decide yourself: Either add an issue for adding a test that checks the stats or add a test directly or ignore this if you think its overdoing it. Thank you.

PR #970, opened 2015-04-23
Hypercube iteration

### LINKED ISSUE #966: Table not found
http://secxbrl.xbrl.io/v1/_queries/public/api/components.jq?_method=POST&format=xml&aid=0001171500-15-000007&token=0fef88d3-01b2-4e73-9521-90750cfbafa2

gislenius:
Fixes #966


queries/public/api/components.jq
```diff
@@ -177,7 +178,8 @@ let $res as object* :=
     order by $component.Archive, $component.Label
     let $archive as object? := $archives[archives:aid($$) eq $component.Archive]
     let $entity as object? := $entities[entities:eid($$) = $archive.Entity]
-    for $found-hypercube allowing empty in components:hypercubes($component)
+    for $found-hypercube allowing empty in
+        hypercubes:hypercubes-for-components($component).Name
```
dknochen: why does it make a difference?

dknochen:
just a question for clarification :)

queries/public/api/components.jq
gislenius: components:hypercubes returns the report element hypercubes (in most cases from the presentation network).
hypercubes:hypercubes-for-components returns the hypercubes from the Hypercubes field (the definition network).
Since the HYPERCUBE-NOT-FOUND error is thrown based on the Hypercubes field, it leads to errors if a hypercube is in the presentation network, but not in the definition network.
So it makes more sense to iterate on the Hypercubes field than on the report elements.
Note that I raised this issue against the infoset generator (https://github.com/28msec/xbrl-infosetgenerator/issues/208)

dknochen: thanks for clarifying :+1:

PR #971, opened 2015-04-23
Num networks is one

gislenius:


queries/public/api/components.jq
```diff
@@ -60,7 +60,6 @@ declare function local:statistics($component as object, $hypercube as string?) a
 {
   {
     NumRules: size($component.Rules),
-    NumNetworks: components:num-networks($component),
```
dknochen: why is it meaningless? should we remove the function alltogether then?

gislenius: Hi Dennis, yes.
Basically, what we call networks is based on the XBRL spec (one network per extended link role/name and arc role/name combination). What Charles calls network is an extended link role. So, there is exactly one network per network (extended link role).
Also, there is no point exposing the XBRL networks to the end user I think, so I just removed it.
Would you like to get rid of the num-networks function?

dknochen: thank you :)

yes, I think we should get rid of the function if we don't expose it. I think we agree that we don't change our semantics of networks, right? We just remove exposing this statistic to prevent confusion.

Do you agree?

gislenius: We don't change our semantics. It's just that remove the statistics. To the outside world, we use the word Network the Charles way. Internally, we use it our way.

PR #997, opened 2015-04-30
Improve test performance.

gislenius:


wcandillon:
very cool!

wcandillon:
yes it is much much much faster!

PR #998, opened 2015-04-30
Transaction time latest (generic replacement for HideAmendedFacts)

### LINKED ISSUE #992: Aggregate sec:Accepted cell by cell rather than entity+period wise.

gislenius:
Fixes #992


gislenius:
http://rendering.secxbrl.info/#?url=http:%2F%2Fsecxbrl-dev8.28.io:80%2Fv1%2F_queries%2Fpublic%2Fapi%2Fspreadsheet-for-component.jq%3Ftoken%3Dfoobar%26ticker%3Dko%26fiscalYear%3DALL%26fiscalPeriod%3DFY%26disclosure%3DBalanceSheet%26format%3Djson%26_token%3DOWJkM1lPRWRiZmxjMjRibjZXWFprMkZBTHJzPToyMDE1LTA1LTA0VDIzOjMwOjE0Ljg5ODYyNFo%3D

gislenius:
@dknochen I'm open to any better suggestions about the option name (AggregateAspectMax).

dknochen:
when comparing two entities:
http://rendering.secxbrl.info/#?url=http:%2F%2Fsecxbrl-dev8.28.io:80%2Fv1%2F_queries%2Fpublic%2Fapi%2Fspreadsheet-for-component.jq%3Ftoken%3Dfoobar%26ticker%3Dt%26ticker%3Dvz%26fiscalYear%3DALL%26fiscalPeriod%3DFY%26disclosure%3DBalanceSheet%26format%3Djson%26_token%3DOWJkM1lPRWRiZmxjMjRibjZXWFprMkZBTHJzPToyMDE1LTA1LTA0VDIzOjMwOjE0Ljg5ODYyNFo%26override%3Dtrue%26validate%3Dtrue

there is a problem with the headers:
image

and facts:
image

should be: 
image

gislenius:
Hi Dennis,

These bugs are not related to this PR and are also on the master. Can you file issues instead so we investigate?

The second bug is because rules should be regenerated when merging components, i.e., the same logics as in the infoset generator to generate roll-ups should be implemented in JSONiq -- that's a long shot and I don't think we'll address this in the short term.

Thanks and kind regards,
Ghislain

queries/modules/io/28/modules/xbrl/facts.jq
```diff
@@ -77,12 +77,9 @@ jsoniq version "1.0";
  : <li><b>Lang</b>: language identifier according to http://www.ietf.org/rfc/rfc3066.txt,
  :     i.e. only return footnotes etc. for this specific language</li>
  : <li><b>AuditTrail</b>: if set to "debug" the audit trails will be more verbose</li>
- : <li><b>facts-for-archives-and-concept</b> (deprecated, use filters instead):
- :     to override how underlying facts are
- :     resolved, for example with finer-grained, profile-specific filtering (option value
- :     must be a function item). facts:facts-for-archives-and-concepts#3 is used by
- :     default, but it is possible to supply another function that, for examples, filters
- :     irrelevant facts out.</li>
+ : <li><b>AggregateAspectMax</b>: an aspect for which only the latest value is retained,
```
dknochen: this is indeed a bit confusing. What about:

1. CurrentVersionOnly
2. MostRecentVersionOnly

In the end, what we are doing is Versioning, isn't it?

queries/modules/io/28/modules/xbrl/profiles/sec/core.jq
```diff
@@ -260,7 +125,8 @@ declare function sec:edgar-facts-for-archives(
             IsExtension: false
           }
         }
-      }
+      },
+      "AggregateAspectMax": "sec:Accepted"
```
dknochen: the option key should be a static variable.

queries/public/api/facts.jq
```diff
@@ -282,11 +280,10 @@ let $facts :=
       },
       { "ConceptMaps" : $map }[exists($map)],
       { "Rules" : [ $rule ] }[exists($rule)],
-      { "Concepts" : $report.Concepts }[exists($report) and exists($report.Concepts)]
+      { "Concepts" : $report.Concepts }[exists($report) and exists($report.Concepts)],
+      { AggregateAspectMax: "sec:Accepted" }[$profile-name eq "sec"]
```
dknochen: should we use fsa:Submitted in case of japan?

queries/public/test/secxbrl/facts.jq
```diff
@@ -132,7 +132,7 @@ local:check({
         fiscalPeriod:"Q1",
         concept:"us-gaap:Assets"
     }),
-    tagconcept: local:test-facttable(30, {
+    tagconcept: local:test-facttable(31, { (: Wrong stamp for AT&T, two FY 2014s:)
```
dknochen: the wrong stamping is not an infoset generator bug. Its reported as FY although its a 10-Q.

queries/public/test/secxbrl/facttable-for-component.jq
```diff
@@ -64,7 +64,7 @@ test:check-all-success({
           disclosure:"BalanceSheet"
       },
       function($b as item*) as item* { count($b.FactTable[]) },
-      153
+      113
```
dknochen: why does it return less facts?

queries/modules/io/28/modules/xbrl/facts.jq
```diff
@@ -1275,7 +1286,7 @@ declare %private function facts:facts-for-direct(
                   return { $open-dimension : "xbrl28:Domain" }
                 else ()
               |}
-              let $removals := seq:value-except(keys($fact.$facts:ASPECTS), $hc-dimension-names)
+              let $removals := ($aggregate, seq:value-except(keys($fact.$facts:ASPECTS), $hc-dimension-names))
```
dknochen: shouldn't this be part of the result? It is still a key aspect. In the facttable or facts endpoint its an important property. Maybe, we should only remove it from the spreadsheed view. But then we should probably add it in the audit-trail. Otherwise, we have no way to trace it to the source fact.

What do you think?

dknochen:
I like the proposal a lot :) its a great improvement. We can discuss the comments if you want...

thanks a lot!

queries/modules/io/28/modules/xbrl/facts.jq
gislenius: DONE (2).

queries/public/api/facts.jq
gislenius: DONE

queries/modules/io/28/modules/xbrl/profiles/sec/core.jq
gislenius: NOT DONE as discussed.

queries/public/test/secxbrl/facttable-for-component.jq
gislenius: Precisely because the most current version is taken! :-D 153 was a bug...

queries/modules/io/28/modules/xbrl/facts.jq
gislenius: If it's okay, I'll add it as an audit trail. I think it's logical to remove this aspect from the output aspects, precisely because it "vanishes" by taking the latest version. It shields users from this aspect, in the same way a version-control database shields user from time when they look at the current version. Makes sense?

queries/public/test/secxbrl/facttable-for-component.jq
dknochen: Sorry, I still don't understand. How can the most current version be a "removed fact" that has existed in a previous version? Shouldn't it be exactly the same number, because if the fact is no longer in reported in the following year, we would automatically still take the one from the previous year, but not remove it?

queries/modules/io/28/modules/xbrl/facts.jq
dknochen: I think adding it to the audit-trail is perfectly sufficient. And, the audit trail looks very good to me. Contains everything needed.

queries/public/test/secxbrl/facttable-for-component.jq
gislenius: That's not what's happening. There are no removed facts. This test goes over two filings. In the 153 facts, there are 40 cell collisions with two values (one old, one overriden), because the old code was not even performing the task of taking the latest values (we just didn't notice it). You can see it here:

http://rendering.secxbrl.info/#?url=http:%2F%2Fsecxbrl.28.io:80%2Fv1%2F_queries%2Fpublic%2Fapi%2Fspreadsheet-for-component.jq%3Ftoken%3Dc3049752-4d35-43da-82a2-f89f1b06f7a4%26ticker%3Dko%26fiscalYear%3DALL%26fiscalPeriod%3DFY%26disclosure%3DBalanceSheet%26format%3Djson%26_token%3DOWJkM1lPRWRiZmxjMjRibjZXWFprMkZBTHJzPToyMDE1LTA1LTA0VDIzOjMwOjE0Ljg5ODYyNFo%3D

This PR solves this and takes the latest value in each cell. It's its purpose (see first link in this PR).

Makes sense?

queries/modules/io/28/modules/xbrl/facts.jq
gislenius: Cool ;-) Glad you like it. I'm glad you forced me to add an audit trail.

queries/public/test/secxbrl/facttable-for-component.jq
dknochen: Thank you for the explanation! It makes sense and I was able to reproduce it. Perfect :).

PR #1003, opened 2015-05-02
Switch MongoDB deployment

wcandillon:


wcandillon:
@gislenius Could you look at the failing test? @fcavalieri and I took a look but we were puzzled by it.

gislenius:
@wcandillon Nothing to be worried about. This test was picking the first component in the collection, which means that it was random. It is no wonder it breaks if the order changed and tests should never be random.
I made the test deterministic by forcing a network URI.

wcandillon:
@gislenius makes a lot of sense, thank you so much!

PR #1028, opened 2015-06-02
Entity xml

gislenius:


gislenius:
This was a real mess. It's now cleaned up. I kept some fields for backwards incompatibility.

queries/public/api/entities.jq
```diff
@@ -63,7 +62,8 @@ let $entities :=
   for $entity in $entities
   return {|
     project($entity, "_id")[$profile-name eq "sec"],
-    { EID: $entity._id }[empty($entity.EIDs)],
+    { ID: $entity._id }[$profile-name eq "sec"],
```
gislenius: That's just to be backwards compatible.

queries/public/api/entities.jq
```diff
@@ -63,7 +62,8 @@ let $entities :=
...
+    { EIDs: [ $entity._id ] }[empty($entity.EIDs)],
```
gislenius: That's to make EIDs consistent with all profiles.

queries/public/api/entities.jq
```diff
@@ -83,6 +83,21 @@ let $entities :=
     trim($entity, "_id")
   |}
 
+let $entities :=
```
gislenius: That's just backwards compatible stuff as well.

queries/public/api/filings.jq
```diff
@@ -90,21 +89,28 @@ let $summaries :=
...
-          NumAbstracts: archives:num-abstracts($archive),
-          NumConcepts: archives:num-concepts($archive)
-      }
+          Profiles: {
+            SEC: {|
+              { Name: "SEC" },
+              filings:summaries($archive)
+            |}
+          },
+          AccessionNumber: $archive._id
```
gislenius: That's just to be backwards compatible.

queries/public/api/filings.jq
```diff
@@ -119,7 +125,7 @@ let $serializers := {
         switch($profile-name)
         case "sec" return
           <Filings>{
```
gislenius: Will be renamed to Archives at some point, but it will need a heads-up to Charlie because it would break.

wcandillon:
@gislenius very cool :+1:

PR #1036, opened 2015-06-08
Bump gulp-ng-annotate version

wcandillon:


ghislainfourny:
@wcandillon It was not uptodate with the master. I pulled it back and it may help.

PR #1051, opened 2015-06-15
Upgrade with new field.

ghislainfourny:


ghislainfourny:
This adds the cellstore-content field, as discussed with William (to make tests orthogonal to the profile).

PR #1053, opened 2015-06-15
Fix bug in total detection.

ghislainfourny:


queries/modules/io/28/modules/xbrl/resolution.jq
```diff
@@ -302,25 +302,36 @@ declare %private function resolution:convert-concept-relationship-node(
...
+    let $order := $network.Order
+    order by $order
     let $concept as string := $network.Name
     let $default-languages as string* := $components.$components:DEFAULT-LANGUAGE
     let $default-languages := if(exists($default-languages)) then $default-languages else "en-US"
     let $concept-metadata := ($components.Concepts[])[$$.Name eq $concept][1]
     let $calculation-network as object* :=
       descendant-objects(networks:networks-for-components-and-short-names($components, $networks:CALCULATION_NETWORK))[$$.Name eq $concept]
-    let $is-total as boolean := exists($calculation-network.To[])
+    let $preceding-siblings as string* := descendant-objects($networks[$$.Order le $order]).Name
```
dknochen: why descendant-objects? Are they not always on the same level?

queries/modules/io/28/modules/xbrl/resolution.jq
```diff
@@ -302,25 +302,36 @@ declare %private function resolution:convert-concept-relationship-node(
...
+    let $roll-up-children as string* := $calculation-network.To[].Name
+    let $is-total as boolean :=
+      exists($roll-up-children) and
+      (
+        every $roll-up-child as string in $roll-up-children
+        satisfies $roll-up-child = $preceding-siblings
```
dknochen: Is it really important that it's preceding? This is making some strong assumptions. What if someone wants to put the total at the top?

queries/modules/io/28/modules/xbrl/resolution.jq
ghislainfourny: No. On the same level, there can also be a roll-up abstract of which the summed concept is a child.

queries/modules/io/28/modules/xbrl/resolution.jq
ghislainfourny: I'm open to discuss that. The idea is very naive: if you put an underline but some of the operands are below, it looks strange. :-)

queries/modules/io/28/modules/xbrl/resolution.jq
dknochen: ok then. makes sense.

queries/modules/io/28/modules/xbrl/resolution.jq
dknochen: I believe the concrete rendering (underlined or not, bold or not, etc.) is an orthogonal task and should not be considered at this stage. This stage is only structural. But, we didn't agree earlier on this and I'm convinced we won't now. Also, its not a big issue. So, I'm fine either way. 

We can fix this later if the use case arises. :)

Do you agree to this summary?

ghislainfourny: Hi Dennis,

I think you're making a good point here. This should be orthogonal. I'll extend the check to all siblings.

Kind regards,
Ghislain

ghislainfourny:
@dknochen Back to you!

dknochen:
:)

ghislainfourny:
Thanks!

PR #1067, opened 2015-06-23
Typed dimensions

ghislainfourny:


wcandillon:
I tried the http://secxbrl-dev10.28.io/v1/_queries/public/api/facts.jq?token=foobar&ticker=ko&concept=us-gaap:Assets&labels=true&format=atom-xml linked with Tableau and that works awesomely :+1:

PR #1072, opened 2015-06-29
Entity report element

ghislainfourny:
Report-elements.jq returns an error in swissxbrl, where the samples have no entity collection yet. We should be more flexible.


dknochen:
@ghislainfourny do you have a particular filing that we could add to test this?

ghislainfourny:
@dknochen no, because in all filings the entities exists. It's just more resilient for swissxbrl that doesn't have entities yet.

Issue #1066, opened 2015-06-23
Labels not displayed correctly in facts query


ghislainfourny:
http://secxbrl.28.io/v1/_queries/public/api/facts.jq?token=foobar&tag=DOW30&concept=us-gaap:Assets&labels=true&format=html&fiscalPeriod=FY&fiscalYear=2014

PR #1075, opened 2015-07-06
Made label retrieval faster and fixed bug

ghislainfourny:
Fixes https://github.com/28msec/cellstore/issues/1066

Now 900ms (was 1.2 s)

http://secxbrl-dev10.28.io/v1/_queries/public/api/facts.jq?token=foobar&tag=DOW30&concept=us-gaap:Assets&labels=true&format=html&fiscalPeriod=FY&fiscalYear=2014


dknochen:
@ghislainfourny could you explain this change to me offline?

ghislainfourny:
That's an easy question. Yes! :-)

queries/public/api/facts.jq
```diff
@@ -36,6 +44,19 @@ declare variable $local:additional-concepts as object* := (
   }
 );
 
+declare function local:resolve-aspect-name-labels(
```
dknochen: comments :)

dknochen:
we could add this as a test http://secxbrl.28.io/v1/_queries/public/api/facts.jq?token=foobar&tag=DOW30&concept=us-gaap:Assets&labels=true&format=html&fiscalPeriod=FY&fiscalYear=2014
checking that all fields of legal entity are filled in

dknochen:
@ghislainfourny performance improvements plus bug fix is always the best :+1:

ghislainfourny:
@dknochen I added the comment and the new test!

queries/public/test/secxbrl/facts.jq
```diff
@@ -182,6 +182,18 @@ local:check({
...
+          "facts",
+          {
+            tag : "DOW30",
+            concept:"us-gaap:Assets",
+            labels: "true",
+            fiscalPeriod: "FY",
+            fiscalYear: "2014"
+          },
+          function($res as item*) as item* { count(distinct-values($res.FactTable[].Labels."dei:LegalEntityAxis")) },
+          1
```
dknochen: This wouldn't have failed even before. I think, we should really count that its 30.

Well, it looks like its 31... Something seems to be fishy with the DOW30 tag?

dknochen:
@ghislainfourny wow, very cool resolution :)

ghislainfourny:
Thanks ;-)

PR #1076, opened 2015-07-06
Add migration query.

ghislainfourny:


dknochen:
@ghislainfourny I'm not sure what I need to do?

PR #1077, opened 2015-07-06
Top skip

ghislainfourny:


queries/public/api/facttable-for-component.jq
```diff
@@ -278,6 +282,12 @@ let $serializers := {
         <title type='text'>Cell store facts</title>
         <updated>{current-dateTime()}</updated>
        {
+            let $next := $res."FactTable@odata.nextLink"
+            where exists($next)
+            return <link rel="next">{$next}</link>,
```
dknochen: Could be simplified to:

    $res."FactTable@odata.nextLink" ! <link rel="next">{$next}</link>

dknochen:
@ghislainfourny very good start :+1:

One more question: How does it work in the html and spreadsheet view? In the html view we could easily add a first and last row containing a next link in case a nextLink is returned. What do you think?

queries/public/api/facttable-for-component.jq
ghislainfourny: True, but more difficult to read. Your choice! ;-)

ghislainfourny:
Thanks!

Regarding the spreadsheet, I would leave it aside for now, that's a lot of work.

Regarding html: I agree. I will do this.

Kind regards,
Ghislain

ghislainfourny:
There you go ;-)

http://secxbrl-dev10.28.io/v1/_queries/public/api/entities.jq?format=html&tag=DOW30&token=foobar&top=5

dknochen:
@ghislainfourny I don't see the change in the link you posted. Also, there is a merge conflict.

PR #1078, opened 2015-07-07
Now requiring token in all endpoints.

ghislainfourny:


dknochen:
@ghislainfourny Should we let users know about this change?

ghislainfourny:
@dknochen I don't think so. It was never meant to be accessible with no token in the first place.

PR #1091, opened 2015-07-09
Add DELETE method to filings.jq.

ghislainfourny:


ghislainfourny:
TODO:
- add admin token to config file
- set up a test database with write access
- have tests (i) insert an archive into the database (ii) test it's here (iii) delete it (iv) test it's no longer here
- test for unauthorized with test token

ghislainfourny:
Needs dropping the error "A sequential query cannot be invoked using GET" in a subsequent 28.io release.

wcandillon:
@ghislainfourny We can discuss this offline. I recommend to have the `DELETE` method handled in a separate query in the meantime.

wcandillon:
Two small comments:
* While we have an automatic rest mapping, maybe `-DELETE` as a suffix is not user friendly enough? (no strong opinon)
* No tests?

wcandillon:
@fcavalieri this seems to be a good candidate for the new URL rewriter infrastructure.

Issue #1071, opened 2015-06-29
Add slice-and-dice control to facts.jq


ghislainfourny:
Add the ability to assign slicers and dicers, that is, for each dimension, say if it's a slicer (hence, hidden from the output), or say at what position it appears (1, 2, etc) in the output.

xbrl:Entity::order=slice&
xbrl28:FiscalYear::order=1&
xbrl:Concept::order=2

or

xbrl:Entity::slice=true&
xbrl28:FiscalYear::dice=1&
xbrl:Concept::dice=2

or
slice=xbrl:Entity&
dice::1=xbrl28:FiscalYear&
dice::2=xbrl:Concept

Hypercube objects can be extended with a new field for each dimension:

Slice: true
or
Dice: 2 (or converting the aspects object to an array).

dknochen:
@ghislainfourny I think this is a very important feature. I really love it. Before we select one of your proposed implementations: I am wondering if a dimension cannot be automatically detected as being a slice (single value) or dice (multiple values). For example, what would happen if multiple entities are selected and the entity dimension is configured as "slice"?

ghislainfourny:
Hi Dennis,

That's a very good question. If we comply with MDX semantics, a slice may only be an MDX tuple, that is, it is not possible to have a slice with multiple values.

However, in our case, we may have heterogeneous entity IDs, so we may want to still accept multiple values as a slice.

Automatic detection, I think, may not make sense for a fact table : the semantics of slicing/dicing is not in terms of one value or single value, but in terms of showing or hiding the aspect.

The automatic detection you are referring to is done in component renderings, that's called auto-slice. It is desirable to simplify a rendering, but in a fact table it may be confusing to see a column automatically disappear just because there was only one value... Don't you think?

We may still add an auto-slice parameter of course. But we can do this later and that's orthogonal to this implementation.

I hope it makes sense.

dknochen:
Thanks @ghislainfourny. It makes sense. What I was thinking was not to automatically hide every slice (slice as in OLAP). Instead, I thought each dimension could have an order and only if a dimension at the top is a slice (i.e. has only one value), it would disappear. But I'm understanding Slice&Dice in the OLAP way.

If slice/dice is not equivalent to the OLAP slice&dice we could cause some confusion. Maybe, we need to disambiguate then and not call it slice / dice after all? Probably, the first option would be closest to the best choice then? Or:

<Axis>::show=true|false
<Axis>::order=1

[...]

PR #1096, opened 2015-07-13
Implement columns.

ghislainfourny:
Implements #1071


dknochen:
Wow, if you ask me: this is perfect :+1: 

What we would need to add next is the documentation here:
https://github.com/28msec/cellstore/tree/master/documentation
where we should explicitly point to MDX. Do you agree?

PR #1114, opened 2015-08-11
Begin implementing adding facts.

ghislainfourny:


queries/modules/io/28/modules/xbrl/facts.jq
```diff
@@ -1913,3 +1913,88 @@ declare %an:sequential function facts:remove-facts-for-archives(
...
+  case exists(values($fact.Aspects)[not $$ instance of atomic])
+  return error(QName("facts:INVALID_INPUT"), "Non-atomic aspect.")
+  case not keys($fact.Aspects) = $facts:CONCEPT
+  return error(QName("facts:INVALID_INPUT"), "Concept missing.")
+  case not keys($fact.Aspects) = $facts:ENTITY
+  return error(QName("facts:INVALID_INPUT"), "Entity missing.")
+  case empty($aid) and not keys($fact.Aspects) = $facts:ARCHIVE
+  return error(QName("facts:INVALID_INPUT"), "Archive missing.")
+
+  case exists(keys(trim($fact.Concept, ("PeriodType", "DataType", "Balance"))))
```
dknochen: shouldn't we fetch the concept from the archive and fill these fields in automatically? And, not allow the user to provide the Concept field?

queries/modules/io/28/modules/xbrl/facts.jq
```diff
@@ -1913,3 +1913,88 @@ declare %an:sequential function facts:remove-facts-for-archives(
...
+ :)
+declare %an:sequential function facts:insert(
+    $fact as object,
+    $archive-or-id as item?,
+    $options as object?) as object*
+{
+  let $aid as atomic? := archives:aid($archive-or-id)
+  let $transaction-time as string? := $options.MostRecentVersionAspect
+  let $current-date as string :=
+      format-date(current-date(), "[Y0001]-[M01]-[D01]")
```
dknochen: I think that the transaction time by default should be a timestamp, don't you think? If you think about someone editing a cell in a spreadsheet the granularity of a day doesn't make sense.

queries/modules/io/28/modules/xbrl/facts.jq
```diff
@@ -1913,3 +1913,88 @@ declare %an:sequential function facts:remove-facts-for-archives(
...
+      else insert json { $transaction-time : $current-date } into $fact.Aspects,
+      if(exists($aid))
+      then
+        if(exists($fact.Aspects.$facts:ARCHIVE))
+        then replace value of json $fact.Aspects.$facts:ARCHIVE
+             with $aid
+        else insert json { $facts:ARCHIVE : $aid } into $fact.Aspects
+      else (),
+      if(not $fact.KeyAspects[] = $transaction-time)
+      then append json $transaction-time into $fact.KeyAspects
```
dknochen: does this work with $transaction-time being an empty sequence?

queries/public/api/add-facts.jq
```diff
@@ -0,0 +1,67 @@
+import module namespace config = "http://apps.28.io/config";
```
dknochen: @fcavalieri Could you point us to an example where the api/facts/POST.jq rewriting is setup. Maybe, we could use it here already?

fcavalieri: Hi @dknochen, I can create a PR for this branch which sets it up :)

queries/public/api/add-facts.jq
```diff
@@ -0,0 +1,67 @@
...
+  (),
+  (),
+  $aid)
+
+let $fact as object? := parse-json($body)
+
+let $archives-not-found as boolean := exists($aid) and empty($archives)
+
+return switch(true)
+       case $archives-not-found return api:not-found("archive")
```
dknochen: that is a security breach. The archives should only be checked after the authorization.

queries/public/test/secxbrl/facts.jq
```diff
@@ -80,6 +81,25 @@ declare %an:nondeterministic function local:test-labels-aids() as item
     return test:assert-eq-array($expected, $actual, $status, test:url($endpoint, $params))
 };
 
+if($config:content[] = "SEC-rw")
+then {
+  mw:insert("archives", { _id: "foobar" });
```
dknochen: this can cause some problems in the parallel test execution. Maybe, we should use $config:frontend-url as _id?

dknochen:
@ghislainfourny This is so cool :+1: Really cool

In addition to the comments, I would like to ask whether we could do the following as well (it would not only add tests for this new feature but also greately help me with my demo):
+ add a new db with read/write
+ add a new config risk-data.jq that connects to this db
+ add tests for adding facts (also multiple facts that overwrite existing facts to test the versioning)

what do you think?

queries/public/api/add-facts.jq
fcavalieri: @dknochen What rewrite would you like?

dknochen: @fcavalieri right now we have:

GET: queries/public/api/facts.jq
POST: queries/public/api/add-facts.jq

I assume that a proposal for a rewrite would be:

GET: queries/public/api/facts.jq -> queries/public/api/facts/GET.jq
POST: queries/public/api/facts.jq -> queries/public/api/facts/POST.jq

Would that work? 

thank you :)

fcavalieri: @dknochen 
you mean /v1/_queries everywhere?

We can do even 
GET http://project.28.io/api/facts -> http://project.28.io/v1/_queries/public/api/facts/GET.jq
POST http://project.28.io/api/facts -> http://project.28.io/v1/_queries/public/api/facts/POST.jq

What do you think?

Best
Federico

dknochen: @fcavalieri 
yes, I meant /v1/_queries everywhere:

GET http://project.28.io/v1/_queries/api/facts -> http://project.28.io/v1/_queries/public/api/facts/GET.jq
POST http://project.28.io/v1/_queries/api/facts -> http://project.28.io/v1/_queries/public/api/facts/POST.jq

I'm not sure if we also need the web API. If its not a big deal, it would be useful, I think.

cheers

fcavalieri: @dknochen 
There is no web API involved in the rewrite:
GET http://project.28.io/api/facts -> http://project.28.io/v1/_queries/public/api/facts/GET.jq
POST http://project.28.io/api/facts -> http://project.28.io/v1/_queries/public/api/facts/POST.jq

The request http://project.28.io/api/facts becomes http://project.28.io/v1/_queries/public/api/facts/GET.jq before sausalito sees it. The only change respect to your proposal is that it is shorter.

queries/modules/io/28/modules/xbrl/facts.jq
ghislainfourny: https://github.com/28msec/cellstore/issues/1116

queries/modules/io/28/modules/xbrl/facts.jq
ghislainfourny: Good point!

queries/modules/io/28/modules/xbrl/facts.jq
ghislainfourny: Fixed.

queries/public/api/add-facts.jq
ghislainfourny: I think it makes sense! :-)

queries/public/api/add-facts.jq
ghislainfourny: I'll raise a bug. It's also in delete-filings.jq and has to be addressed all at once.

ghislainfourny: https://github.com/28msec/cellstore/issues/1117

queries/public/test/secxbrl/facts.jq
ghislainfourny: DONE

queries/public/api/add-facts.jq
dknochen: Ah, you mean also adding: 

http://project.28.io/v1/_queries/public/api/facts.jq -> http://project.28.io/v1/_queries/public/api/facts/<method>.jq

Allowing urls with and without the ".jq"? Good point.

dknochen:
image

one failing XQlint

queries/public/api/add-facts.jq
dknochen: I think that @fcavalieri will do the renaming in his MP anyway ;)

fcavalieri: Hello @dknochen , @ghislainfourny ,
I am getting confused with your requirements. Can you explain them to me?
Some of the things you are asking for are not currently allowed.

To start with the same set of information:
* If a user request is a project API call it cannot (it is not allowed, not impossible) be rewritten. Anything that start with <project-endpoint>/v1/, roughly. That is http://project.28.io/v1/_queries/public/api/facts.jq cannot become http://project.28.io/v1/_queries/public/api/facts/*method*.jq. It can be allowed if we think it makes sense, either user-customizable, or just with the query.jq /query/%{METHOD}.jq rewrite. 

* On the other hand, anything that does not start with <project-endpoint>/v1/ can be rewritten completely freely. e.g.: http://project.28.io/api/facts can be transformed into any request to the same project, e.g.: http://project.28.io/api/facts/%{REQUEST_METHOD}.jq or  http://project.28.io/v1/_queries/api/facts/%{REQUEST_METHOD}.jq.
In the first case the request becomes a web api call (legacy code), in the second one a call to the _queries API.

* If a request is rewritten, the original one never reaches sausalito. Only the rewritten request does, as if the user made the rewritten request directly.

dknochen: @fcavalieri I didn't know these limitations. Thanks for explaining. Rewriting the web api should be sufficient to start with then. Within the _queries api the rewrite could probably be done automatically (without requiring the user to configure it):

Request: http://project.28.io/v1/_queries/public/api/facts
1. public/api/facts/%{METHOD}.jq exists? then rewrite to public/api/facts/%{METHOD}.jq
2. public/api/facts.jq exists? then rewrite to public/api/facts.jq

Is that what you suggested as well?

dknochen:
@ghislainfourny some tests are failing:
image

PR #1115, opened 2015-08-20
:pencil: Add risk-data db.

ghislainfourny:


dknochen:
@ghislainfourny thank you for helping with this!

